PR #3, opened 2022-02-03
Simplify core

matheusps:
### How it was

Once upon a time (haha)... We tried to create a core that would be capable of handling different style technologies & theme structures, all of this, selected by the developer. For the theme consumption, we defined the `plugins` that any user could define how a theme is consumed. For example:

Imagine that the theme object is:

```js
const theme = {
  bg: {
    primary: 'blue'
  }
}
```

We want to consume the `bg` entry from the theme, which can only be consumed by the `background` CSS property, that can use `bg` as an alias. In other words, the:

```js
const csx = {
  bg: 'primary'
}

// or
const csx = {
  background: 'primary'
}
```

should have the output:

```js
const output = {
  background: 'blue'
}
```

The `bgPlugin` defined this behavior using the code above:

```js
const bgPlugin = createPlugin({
  namespaces: ['bg'],
  aliases: {
    bg: 'background'
  },
  rules: {
    background: 'bg'
  }
})
```

As you may imagine, we developed a set of plugins to handle how CSS properties access our theme values. We also needed to create the `runtime` abstraction, which creates an instance of every plugin in admin-ui. This approach has some problems:

1. It is not desirable for users to change how the theme is consumed (update in the product specs). 
2. It assumes that everything is dynamic, so we need to run more js code in runtime (it harms FCP and LCP metrics). 
3. It is hard to maintain since we must care about multiple plugin instances, validation, conflicts and build.

So, in the end, the problem that we WANTED to solve is not the problem we SHOULD solve. The current structure is overkill for the specs. 

### What's changed?

- Static theme.
- Straightforward style consumption with `styles` function.
- No plugins, runtime, and system abstractions. They were replaced by a single `createCsx` function, to join the emotion instance with the `styles` function.
- Less runtime code, since we don't need to instance plugins anymore.

### There are any breaking changes?

Just if you were using the `unstableCreateAdminUI` function (this is not the reality for most of our users). To use the same behavior is you must use the `__theme` prop.

#### Before

```js
import { theme, merge,  createSystem, unstableCreateAdminUI } from '@vtex/admin-ui'

const customTheme = {}
const unstableAdminUI = unstableCreateAdminUI(merge(theme, customTheme))

const [ThemeProvider] = createSystem({
  key: 'example',
  unstableAdminUI
})
```

#### After

```js
import { theme, merge,  createSystem } from '@vtex/admin-ui'

const customTheme = {}

const [ThemeProvider] = createSystem({
  key: 'example',
  __theme: merge(theme, customTheme)
})
```

#### Types of changes


packages/admin-ui-core/src/styles.ts
```diff
@@ -0,0 +1,132 @@
...
+    for (let i = 0; i < value.slice(0, mediaQueries.length).length; i++) {
+      const media = mediaQueries[i]
+
+      if (!media) {
+        next[key] = value[i]
+        continue
+      }
+
+      next[media] = next[media] ?? {}
+      if (value[i] == null) continue
```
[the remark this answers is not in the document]
Lenderson: yes, make sense now to me 😉. 
**Suggestion:** Maybe break a line before and after the conditional block or add curly braces could be good in this case to make reading easier.
```suggestion
     next[media] = next[media] ?? {}
     if (value[i] == null) {
         continue
     }
     (next[media] as Record<string, unknown>)[key] = value[i]
```

packages/admin-ui-core/src/styles.ts
```diff
@@ -0,0 +1,129 @@
...
+function resolveResponsiveArray(
+  styles: Exclude<StyleProp, ThemeDerivedStyles>
+) {
+  const next: Exclude<StyleProp, ThemeDerivedStyles> = {}
+
+  for (const k in styles) {
+    const key = k as keyof typeof styles
+    const value = styles[key]
+
+    if (value == null) continue
```
Lenderson, line 43: **Suggestion:** If you to check if value is `null` or not it could be better use `value === null`, because `undefined == null` is `true`, but if you also want to consider value as undefined, maybe  `!!value` could it be a better option.

matheusps: Tried, but. The logic changes. So, I'm keeping that. The truth about this function is that we must re-write it. I know the logic is not perfect right now. But, I prefer keeping that because I'm refactoring. In another PR I can upgrade the logic around it. What do you think?

packages/admin-ui-core/src/styles.ts
```diff
@@ -0,0 +1,132 @@
...
+      const media = mediaQueries[i]
+
+      if (!media) {
+        next[key] = value[i]
+        continue
+      }
+
+      next[media] = next[media] ?? {}
+      if (value[i] == null) continue
+      ;(next[media] as Record<string, unknown>)[key] = value[i]
```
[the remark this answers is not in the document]
matheusps: Removed the need of caching. It solved.

packages/admin-ui-core/src/styles.ts
```diff
@@ -0,0 +1,135 @@
...
+import { resolveRule } from './rules'
+import { canSplit, split } from './splits'
+import { createTransform } from './transforms'
+import type { StyleObject, StyleProp, ThemeDerivedStyles } from './types'
+import { theme as defaultTheme } from './theme'
+
+/**
+ * Available system breakpoints
+ */
+const breakpoints = ['40em', '48em', '64em', '75em']
```
lucasaarcoverde, line 15: Should it be removed since we already have the `./src/tokens/breakpoints.ts`?

lucasaarcoverde (approved):
LGTM, just two comments

Lenderson (approved):

packages/admin-ui-core/src/styles.ts
Lenderson: I agree, make sense ;)

PR #109, opened 2022-03-29
feat: create documentation footer

Lenderson:
Create documentation [footer](https://www.figma.com/file/hcO2T1NcrgGZelG6TvAtKn/Documentation?node-id=831%3A264431)

Check in the preview link. At the current moment, we are not showing the links footer as **Privacy**, **Terms of use** and others because we do not have all the internal links, so @davicostalf asked to not show them now, **but the footer is already prepared for showing those links in the future.**

#### Types of changes
- [x] Trivial change (small fix or feature that doesn't impact functionalities)

### DIFF
--- a/packages/admin-ui-docs/src/theme/Footer/index.js
+++ b/packages/admin-ui-docs/src/theme/Footer/index.js
@@ -1,0 +1,109 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import React from 'react';
+import clsx from 'clsx';
+import Link from '@docusaurus/Link';
+import { useThemeConfig } from '@docusaurus/theme-common';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import styles from './styles.module.scss';
+import ThemedImage from '@theme/ThemedImage';
+
+function FooterLink({ to, href, label, prependBaseUrlToHref, ...props }) {
+  const toUrl = useBaseUrl(to);
+  const normalizedHref = useBaseUrl(href, {
+    forcePrependBaseUrl: true,
+  });
+  return (
+    <Link
+      {...(href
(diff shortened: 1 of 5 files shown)


kevinch:
I think this is missing the left side links. Or is that on purpose?

packages/admin-ui-docs/src/theme/Footer/index.js
```diff
@@ -0,0 +1,108 @@
...
+      )}>
+      <div className={styles.leftContent}>
+        <FooterLogo
+          alt={logo.alt}
+          sources={sources}
+          width={logo.width}
+          height={logo.height}
+        />
+        <span className={styles.name}>
+          Admin UI <span className={styles.version}>v.01</span>
```
kevinch, line 89: Hard coded version number? Could this be dynamic?

packages/admin-ui-docs/src/theme/Footer/styles.module.scss
```diff
@@ -0,0 +1,98 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+.footer {
+  position: relative;
+  padding: 48px 24px;
+  background: #F8F8F8;
```
kevinch, line 10: Could this be a token? Not sure if it makes sense here.

Lenderson: Our tokens only work in a object style, I could try to create a object style for this docusaurus component and see if make sense. 🤔

packages/admin-ui-docs/src/theme/Footer/index.js
Lenderson: After talk to @davicostalf I am removing the version number because we should think a better way of showing this and right now we are not making any distinction of using Admin UI by its version.

lucasaarcoverde (approved):
LGTM

Issue #37, opened 2022-02-08
Modal doesn't call the onClose on click outside (hideOnClickOutside) or manual calls to `modal.hide`


lucasaarcoverde:
# Bug Report

- [x] I've read and understood the [Contributing guidelines](https://admin-ui-docs.vercel.app/engineering/contributing/) and have done my best effort to follow them.
- [x] I've searched for any related issues and avoided creating a duplicate issue.

## What happened

1. When you click outside of the `Modal`, the `onClose` isn't called:

```tsx
// Doesn't call the onClose on click outside case
<Modal state={state} onClose={handleClose} hideOnClickOutside />
```

2. When you call `state.hide` manually it doesn't call the `onClose`:

```tsx
const handleRefresh = () => {
  state.hide() // // Doesn't call the onClose passed to Modal
}

<Modal state={state} onClose={handleClose} hideOnClickOutside />
```

## What I expected to happen

It should call `onClose` in both situations.

## Reproducing

- Admin-ui version in your `package.json`: 0.114.3
- Live link with error:

Code
```tsx
import {
  createSystem,
  Modal,
  ModalHeader,
  useModalState,
  ModalDisclosure,
  ModalContent
} from "@vtex/admin-ui";
import { useEffect } from "react";

import { render } from "react-dom";

const [SystemProvider] = createSystem({
  key: "codesandbox-example"
});

function App() {
  const state = useModalState();

  const handleClose = () => {
    console.log("handleClose");
  };

  useEffect(() => {
    if (state.visible) {
      const timeoutId = setTimeout(() => state.hide(), 5000);
      return () => clearTimeout(timeoutId);
    }
  }, [state.visible]);

  return (
    <>
      <ModalDisclosure state={state}>
        <button>Open Modal</button>
      </ModalDisclosure>

      <Modal state={state} aria-label="modal-example" onClose={handleClose}>
        <ModalHeader title="Title" />
        <ModalContent>Example</ModalContent>
      </Modal>
    </>
  );
}

render(
  <SystemProvider>
    <App />
  </SystemProvider>,
  document.getElementById("root")
);
```

Steps to reproduce:

1. Click in `Open Modal`
2. Click outside of the modal

RESULT: It will close the modal, but isn't gonna call the `onClose` event

---

1. Click in `Open Modal`
2. Wait for 5 seconds (it will close calling the `state.hide`)

RESULT: It will close the modal, but isn't gonna call the `onClose` event

[...]

PR #115, opened 2022-03-31
[AU-117] Calls onClose on state.hide

gtkatakura:

- Associate the `Modal#onClose` with the `modalState.hide`, so when someone calls it, it will call the `onClose` passed to the `Modal`.

- Enable people to use `modal.hide` but don't forget to call the `onClose` event when the modal closes.

- Just run `yarn test`

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/components/Modal/state.internals.tsx
+++ b/packages/admin-ui/src/components/Modal/state.internals.tsx
@@ -1,0 +1,28 @@
+import { useEffect } from 'react'
+import { useCallbackRef } from '@vtex/admin-ui-hooks'
+import type { ModalStateReturn } from './state'
+
+export const kListeners = Symbol('useModalState#listeners')
+
+export type ModalStateListeners = {
+  close: Set<() => void>
+}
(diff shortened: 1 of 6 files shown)


packages/admin-ui/src/components/Modal/state.internals.tsx
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react'
+import { useCallbackRef } from '../../../../admin-ui-hooks/src/useCallbackRef'
+import type { ModalStateReturn } from './state'
+
+export const kListeners = Symbol('useModalState#listeners')
```
gtkatakura, line 5: I don't want to expose the `kListeners` to avoid people using these details implementation, so I created this new file to be used just internally by the library. Not sure which kind of patterns do the team use to define private vs public stuff.

matheusps:
The onClose doesn't exist anymore since the modal was updated. We will resolve https://github.com/vtex/admin-ui/issues/37 in other way

PR #124, opened 2022-04-07
feat(button): add codemod

matheusps:

Add the `admin-ui-codemod` package to store all codemods that we want to share. The first one is the `button-review`.

Codemods are useful to speed up the migration process if the tasks are repetitive. 

- Clone the repo
- Navigate to `packages/admin-ui-codemod`
- Use `npm link`
- In any folder use the `admin-ui-codemod` script

Once this is published, the `npx` script will work, for example:

```bash
$ npx admin-ui-codemod button-review <path>
```

#### Types of changes

### DIFF
--- a/packages/admin-ui-codemod/src/button-review.js
+++ b/packages/admin-ui-codemod/src/button-review.js
@@ -1,0 +1,60 @@
+function replace(source, j, componentName) {
+  return j(source)
+    .find(j.JSXOpeningElement, { name: { name: componentName } })
+    .forEach((p) => {
+      const { name, attributes, selfClosing } = p.value
+
+      const newAttributes = attributes.map((attribute) => {
+        if (attribute.name.name === 'size') {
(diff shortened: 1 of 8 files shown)


packages/admin-ui-codemod/src/button-review.js
```diff
@@ -0,0 +1,60 @@
...
+    regular: 'normal',
+  }[value]
+}
+
+function transformVariant(value) {
+  if (!value) return 'primary'
+
+  return {
+    'adaptative-dark': 'neutralTertiary',
+    'adaptative-light': 'neutralTertiary',
```
lucasaarcoverde, line 44: I'm not sure if someone is using this, but IMO we shouldn't add a replacement to this variant because it doesn't have an equivalent. Instead, we should let it break so the user can decide how to approach this

lucasaarcoverde (approved):
LGTM, just a suggestion

PR #137, opened 2022-04-12
Radio & RadioGroup component review

lucasaarcoverde:

Implement design specs changes.

[Figma File](https://www.figma.com/file/sepBhrHViUHyrTnktkuHaW/Admin-UI-%E2%80%93-Components?node-id=1270%3A312852)

#### Types of changes

## Release Note

Below is a summary of the main changes in these components, for detailed information check the Radio documentation.

### Radio

| Property | Changes                                                                                                                              |
| -------- | ------------------------------------------------------------------------------------------------------------------------------------ |
| size     | This property no longer exists, now the Radio only has a unique size.                                                                |
| state    | There is no need to pass the state returned by the `useRadioState` hook to the Radio anymore. You can pass only to the `RadioGroup`. |
| label    | Now the Radio has a built-in implemented label, to add it just use the `label` property.                                             |
| helpText | Use this property to add a help text to the Radio.                                                                                 |

### RadioGroup

| Property    | Changes                                                                                                          |
| ----------- | ---------------------------------------------------------------------------------------------------------------- |
| orientation | This property no longer exists, you must use the `direction` instead                                               |
| helpText    | Adds a help text to the RadioGroup.                                                                            |
| errorText   | Adds an error description below the RadioGroup indicating an error message. It appears when `error` is set to true. |
| error       | Defines whether the RadioGroup is with an error or not.                                                          |
| optional    | Defines whether the RadioGroup is an optional field or not.                                                      |

**Before**

```jsx
function Example() {
  // with a default value: const state = useRadioState({ state: value })
  const state = useRadioState()

  return(
    <RadioGroup state={state} label={label}>
      <Label>
        <Radio state={state} value={value} />
      </Label>
    </RadioGroup>
  )
}
```

**After**

```jsx
function Example() {
  // with a default value: const state = useRadioState({ defaultValue: value })
  const state = useRadioState()

  return (
    <RadioGroup state={state} label={label}>
      <Radio label={label} value={value} />
    </RadioGroup>
  )
}
```


packages/admin-ui/src/radio/Radio.style.ts
```diff
@@ -0,0 +1,85 @@
...
+import { get } from '@vtex/admin-ui-util'
+
+export const checkmark = style({
+  content: '""',
+  position: 'absolute',
+  display: 'block',
+  borderRadius: 'circle',
+  top: '0.3rem',
+  size: '0.5rem',
+  bg: '$form.neutral',
```
matheusps, line 11: This token is weird to me because you are switching between tones.

packages/admin-ui/src/radio/Radio.style.ts
```diff
@@ -0,0 +1,85 @@
...
+})
+
+export const baseline = style({
+  ...focusVisible('neutral'),
```
matheusps, line 15: Shouldn't be main?

packages/admin-ui/src/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export default {
+  title: 'admin-ui-review/radio',
+  component: Radio,
+} as Meta
+
+export const Playground: Story<RadioProps> = (args) => {
+  const state = useRadioState()
+
+  return <Radio {...args} state={state} />
```
matheusps, line 15: Could it also be a controlled component that behaves just like a regular radio?

packages/admin-ui/src/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import type { RadioProps } from './index'
+import { Radio, useRadioState } from './index'
+
+export default {
+  title: 'admin-ui-review/radio',
+  component: Radio,
+} as Meta
+
+export const Playground: Story<RadioProps> = (args) => {
```
matheusps, line 12: This story is not okay. Even to demonstrate, we should not have a single radio on the screen. I think you should review the radio-group component

kevinch (approved):

packages/admin-ui/src/radio/Radio.stories.tsx
kevinch: Good catch.

packages/admin-ui/src/radio/Radio.style.ts
lucasaarcoverde: Same reason: https://github.com/vtex/admin-ui/pull/137#discussion_r849492849

packages/admin-ui/src/radio/radio.style.ts
```diff
@@ -68,28 +64,21 @@ export const checked = css({
...
-export const small = css({
-  size: '1rem',
-  padding: 1,
-  ':after': {
-    size: '0.375rem',
-  },
+export const labelStyle = style({
+  cursor: 'pointer',
+  display: 'inline-flex',
+  '> span': { marginLeft: '$m' },
```
matheusps, line 81:
```suggestion
  '* + *': { marginLeft: '$m' },
```

packages/admin-ui/src/radio/radio.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as style from './radio.style'
+
+export const Radio = createComponent<typeof RadioButton, RadioOptions>(
+  (props) => {
+    const { label, ...radioButtonProps } = props
+
+    return useElement('label', {
+      baseStyle: style.labelStyle,
+      children: (
+        <>
```
matheusps, line 15:
```suggestion
        <Inline hSpace="$m" noWrap align="center">
```

packages/admin-ui/src/radio/radio-group.tsx
```diff
@@ -0,0 +1,85 @@
...
+    ...radioContainerProps
+  } = props
+
+  const error = initialError && errorMessage
+  const message = error ? errorMessage : helperText
+
+  return useElement('div', {
+    baseStyle: style.radioGroupStyles,
+    children: (
+      <>
```
matheusps, line 29:
```suggestion
      <Stack direction={direction}>
```

packages/admin-ui/src/radio/radio-group.tsx
```diff
@@ -0,0 +1,85 @@
...
+    errorMessage,
+    error: initialError = false,
+    children,
+    ...radioContainerProps
+  } = props
+
+  const error = initialError && errorMessage
+  const message = error ? errorMessage : helperText
+
+  return useElement('div', {
```
matheusps, line 26:
```suggestion
  return useElement('fieldset', {
```

packages/admin-ui/src/radio/radio-group.tsx
```diff
@@ -0,0 +1,85 @@
...
+    baseStyle: style.radioGroupStyles,
+    children: (
+      <>
+        <Text as="span" variant="detail" tone="secondary">
```
matheusps, line 30:
```suggestion
        <Text as="legend" variant="detail" tone="secondary">
```

packages/admin-ui/src/radio/radio-group.tsx
```diff
@@ -0,0 +1,85 @@
...
+          {label}
+        </Text>
+        <RadioContainer {...radioContainerProps}>{children}</RadioContainer>
```
matheusps, line 33:
```suggestion
        <AriakitRadioGroup state={state}><Stack direction="direction" space="$m">{children}</Stack></AriakitRadioGroup>
```

packages/admin-ui/src/radio/radio-group.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <tag.div>
+          {message ? (
+            <Text variant="detail" tone={error ? 'critical' : 'secondary'}>
+              {message}
+            </Text>
+          ) : null}
```
matheusps, line 39: I would create a Message component. Somthing like:
```jsx
function Message(props){
  const hasError = props.error &&  props.errorMessage
  const message = props.error ? props.errorMessage : props.helperText

  return <Text {...} />
}
```

packages/admin-ui-core/src/tokens/border.ts
```diff
@@ -22,7 +22,10 @@ export const border = {
     criticalHover: borderStyles('red50'),
     criticalFocus: borderStyles('red60'),
 
-    mainChecked: borderStyles('blue40'),
-    mainCheckedHover: borderStyles('blue50'),
+    control: borderStyles('gray30'),
+    controlHover: borderStyles('gray50'),
+    controlChecked: borderStyles('blue40'),
+    controlFocus: borderStyles('gray60'),
```
matheusps, line 28: Why this is not on fg and bg (like on the buttons, for ex)?

packages/admin-ui-docs/docs/guidelines/foundations/writing.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+Common writing decisions that you must know when developing or designing Admin UI.
+
+## Invalid Messaging
+
+Represented by the [Critical Tone](/guidelines/foundations/colors#critical), an invalid message alerts people of a validation problem that has occurred and informs them what to do next.
+
+It appears in Forms fields after someone has taken an action. The message draws the user's attention to what has happened and provides guidance to move forward. Optionally, they can also include why something happened and whether the problem will occur again.
+
+### Usage
```
lucasaarcoverde, line 20:
```suggestion
### Code Usage
```

packages/admin-ui-core/src/tokens/border.ts
lucasaarcoverde: Actually, this token shouldn't exist. The way we apply focus on buttons and controls is by using the focus ring (`focusVisible` function)

I added this by mistake, I'll remove it

packages/admin-ui/src/radio/radio.style.ts
```diff
@@ -0,0 +1,84 @@
...
+export const checked = style({
+  ...focusVisible('main'),
+  bg: '$form.controlChecked',
+  color: '$form.controlChecked',
+  border: '$form.controlChecked',
+  position: 'relative',
+
+  '&:after': {
+    ...checkmark,
+    bg: (theme) => get(theme, 'fg.form.controlChecked', ''),
```
marcelovicentegc, line 60: Just for context, why is this color gotten from style's callback `theme`?

marcelovicentegc (approved):

packages/admin-ui/src/radio/radio-group.tsx
```diff
@@ -0,0 +1,80 @@
...
+      error = false,
+      ...restProps
+    } = props
+
+    const orientation = direction === 'row' ? 'horizontal' : 'vertical'
+
+    return useElement('fieldset', {
+      ...restProps,
+      children: (
+        <Set spacing="$l" orientation="vertical">
```
Lenderson, line 29: Just worried with the `Set` component, because Matheus is changing to Stack and probably this would require updates here.

packages/admin-ui/src/radio/radio.style.ts
lucasaarcoverde: We use `fg` tokens in the `color` property since this prop applies color to foreground elements (text, icons, etc.) and we use `bg` tokens in the `background-color` property because this prop applies color to the background. Saying that, you can think of the checkmark as a foreground element in theory, but in practice, we create it by using the `:after` selector, so we have to apply an `fg` token to the `background-color`. 

Let me know if something isn't clear in the explanation and I can try explain it better

marcelovicentegc: Got it! Makes sense now

anitavincent (approved):

packages/admin-ui/src/radio/radio-group.tsx
lucasaarcoverde: done @Lenderson

marcelovicentegc:
@lucasaarcoverde, nitpick: it looks like the checkmark is a little bit off, and the checked styles do not match the Figma specs ([while checked and on focus, the component should have the same background color as while being hovered upon](https://www.figma.com/file/sepBhrHViUHyrTnktkuHaW/Admin-UI-%E2%80%93-Components?node-id=478%3A105988))

Related thread: https://vtex.slack.com/archives/G01HXJ7Q6JZ/p1651000600002559

https://user-images.githubusercontent.com/33183880/165544358-45b691f8-e270-480a-9366-470b4a9ec1d7.mov

marcelovicentegc (changes requested):
Per comment above

lucasaarcoverde:
> @lucasaarcoverde, nitpick: it looks like the checkmark is a little bit off, and the checked styles do not match the Figma specs ([while checked and on focus, the component should have the same background color as while being hovered upon](https://www.figma.com/file/sepBhrHViUHyrTnktkuHaW/Admin-UI-%E2%80%93-Components?node-id=478%3A105988))
> 
> Related thread: https://vtex.slack.com/archives/G01HXJ7Q6JZ/p1651000600002559
> 
>  Screen.Recording.2022-04-27.at.11.40.16.mov

It looks like this only happens on the documentation site. In the storybook, it seems centralized. @Lenderson do you have any clue on this?

Doc
Captura de Tela 2022-04-27 às 15 28 34

Storybook
Captura de Tela 2022-04-27 às 15 28 49

marcelovicentegc:
@lucasaarcoverde, mind sharing a storybook link?

lucasaarcoverde:
> @lucasaarcoverde, mind sharing a storybook link?

I removed the padding and I think it fixed this issue. For some reason, it was changing the behavior only in the documentation site

[Storybook](https://radio-preview.surge.sh/?path=/story/admin-ui-review-radio--radio-group-playground) [Documentation](https://radio-docs.surge.sh/guidelines/components/radio)

packages/admin-ui-core/src/tokens/fg.ts
```diff
@@ -49,5 +49,8 @@ export const fg = {
   form: {
     neutral: colors.black,
     neutralChecked: colors.white,
+
```
kevinch, line 52: Why the line break?

Lenderson (approved):

PR #174, opened 2022-05-03
feat(search): add Search component review

lucasaarcoverde:

As the title says.

[Figma link](https://www.figma.com/file/sepBhrHViUHyrTnktkuHaW/Admin-UI-%E2%80%93-Components?node-id=656%3A286735)

#### Types of changes
- [x] Requires change to documentation, which has been updated accordingly

## 🚨 Breaking Changes 🚨

The way of using the search component is a little bit different. Now we focus on the native usage of input instead of passing the state props directly to the search component. This way, the state hooks can be more like a utility than a must-have, turning the component more flexible and consistent.

**Before**
```jsx
function Example() {
  const state = useSearchState()

  return (
      <Search
        id="search"
        state={state}
        placeholder="Search for a product, category or brand"
      />
  )
}
```

```ts
interface UseSearchStateParams {
  /**
   * Search initial value
   */
  initialValue?: string
  /**
   * Whether is initially loading or not
  */
  initiallyLoading: boolean
  /**
   * Value set in the clean action
   */
  defaultValue?: string
  /**
   * Debounce timeout in ms
   */
  timeoutMs?: number
  /**
   * Action to take on form submition
   */
  onSubmit?: (e: FormEvent<HTMLFormElement>) => void
}
```

```ts
export interface SearchFormState {
   * Debounced search input value
   */
  debouncedValue: string
  /**
   * Sets the value state
   */
  setValue: (value: string) => void
  /**
   * Search input value
   */
  value: string
  /**
   * On clear button is clicked event
   */
  clear: () => void
  /**
   * Whether should show the clear option or not
   */
  showClear: boolean
  /**
   * Whether is loading or not
  */
  loading: boolean
  /**
   * Sets the loading state
  */
  setLoading: React.Dispatch<React.SetStateAction<boolean>>
}

```

**After**
The main difference is that now you can easily use the Search without the utility state hook, like a native input usage. Another thing is that you don't need to pass the `placeholder` anymore, we do that internally, and is already localized.

```jsx
function Example() {
  const { value, onChange, onClear } = useSearchState()

  return (
      <Search
        value={value}
        onChange={onChange}
        onClear={onClear}
      />
  )
}
```

You can also use the `getInputProps` function and spread it to the Search component, since it returns all the essential to the search works correctly.

```jsx
function Example() {
  const { getInputProps } = useSearchState()

  return (
      <Search {...getInputProps()} />
  )
}
```

```ts
export interface UseSearchStateParams {
  /**
   * Search initial value
   */
  initialValue?: string
  /**
   * Search default value. When clicking the clear button the value should be reseted to the default one
   * @default ''
   */
  defaultValue?: string
  /**
   * Debounced value timeout in milliseconds
   * @default 250
   */
  timeout?: number
  /**
   * Search input change handler event
   */
  onChange?: ChangeEventHandler
}
```

```ts
export interface SearchFormState {
  /**
   * Returns the search input props: value, onChange, onClear
   */
  getInputProps: () => GetInputPropsReturn
  /**
   * Debounced search input value
   */
  debouncedValue: string
  /**
   * Sets the value state
   */
  setValue: (value: string) => void
  /**
   * Search input value
   */
  value: string
  /**
   * On clear button is clicked event
   */
  onClear: () => void
  /**
   * Search input change handler event
   */
  onChange: ChangeEventHandler<HTMLInputElement>
}
```


packages/admin-ui-docs/docs/guidelines/components/search.mdx
```diff
@@ -85,11 +184,13 @@ function Example() {
...
-### Persisted state in query params
-
-You can persist the search value within the browser's query params using the `useQuerySearchState` hook, instead of `useSearchState`. The application should be wrapped by `QueryStateProvider`.
+</PageSection>
 
-_watch how your url changes (Note that the persisted value is the debounce value)_
+<PageSection
+  id={tabToc.overview[5].id}
+  title={tabToc.overview[5].value}
+  explanation="You can persist the search value within the browser's query params using the `useQuerySearchState` hook, instead of `useSearchState`. The application should be wrapped by `QueryStateProvider`. Watch how your url changes (Note that the persisted value is the debounce value)."
```
kevinch, line 192:
```suggestion
  explanation="You can persist the search value within the browser's query params using the `useQuerySearchState` hook, instead of `useSearchState`. The application should be wrapped by `QueryStateProvider`. Watch how your url changes (note that the persisted value is the debounced value)."
```

packages/admin-ui/src/search/search.i18n.ts
```diff
@@ -0,0 +1,12 @@
+export const messages = {
+  'en-US': { search: 'Search', clearButton: 'Clear search input' },
```
kevinch, line 2: Is there an issue with the localization team for this file? How is this going to scale in the future?

packages/admin-ui-docs/docs/guidelines/components/search.mdx
```diff
@@ -1,66 +1,169 @@
...
 }
 ```
 
-### Debounce
+</PageSection>
 
-A common UX pattern is to filter the collections (or make a query) while the user types the input. To do that performatively, we often use debounced values. The good thing here, it that useSearchState stores a debounced value for you!
+<PageSection
+  id={tabToc.overview[4].id}
+  title={tabToc.overview[5].value}
```
anitavincent, line 164: Wrong title here

packages/admin-ui-docs/docs/guidelines/components/search.mdx
```diff
@@ -119,59 +220,123 @@ function Example() {
...
+  id={tabToc.codeReference[3].id}
+  title={tabToc.codeReference[3].value}
+>
+
+| Name             | Type                       | Description                       | Required | Default |
+| ---------------- | -------------------------- | --------------------------------- | -------- | ------- |
+| initialValue     | `string`                   | Initial input value               | 🚫       | `''`    |
+| defaultValue     | `string`                   | Value set in the clean action     | 🚫       | `''`    |
+| initiallyLoading | `boolean`                  | If is initially loading           | 🚫       | `false` |
+| onSubmit         | `FormEventHandler<'form'>` | Action to take on submit the form | 🚫       | -       |
```
anitavincent, line 274: "on form submission" or "upon form submission"

packages/admin-ui/src/search/search.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const {
+    state,
+    role = 'search',
+    'aria-label': ariaLabel = label,
+    disabled = false,
+    ...formProps
+  } = props
+
+  const iconName = state.loading ? 'spinner' : 'search'
```
anitavincent, line 32: Just a comment not a change request I'm not sure what's the best approach... I'm always afraid of using string values for consts as these. But maybe typescript makes it safe, as the object won't accept other strings as key

lucasaarcoverde: Yes, in this case, typescript makes it safe. I think it's worth discussing it, I'll let it this way, but if you want you can bring a discussion so the entire team can give an opinion about it

anitavincent (approved):

packages/admin-ui/src/search/search.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <Center csx={styles.innerContainer('end')}>
+            <Button
+              aria-label={formatMessage('clearButton')}
+              icon={<IconXCircle />}
+              variant="neutralTertiary"
+              csx={styles.clearButton}
+              onClick={state.clear}
+            />
+          </Center>
+        ) : null}
```
marcelovicentegc, line 63: Is there a reason for not using a logical AND instead?
```suggestion
        {hasClearButton && (
          <Center csx={styles.innerContainer('end')}>
            <Button
              aria-label={formatMessage('clearButton')}
              icon={<IconXCircle />}
              variant="neutralTertiary"
              csx={styles.clearButton}
              onClick={state.clear}
            />
          </Center>
        )}
```

lucasaarcoverde: It's just a preference because using the AND operator doesn't return the expected result always. [You can see this blog post by Kent C. Dodds for detailed info](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx)

packages/admin-ui/src/search/search.tsx
lucasaarcoverde: I've changed to return the component instead of the strings because it was necessary to add a different style when the search is disabled

packages/admin-ui/src/search/search.tsx
marcelovicentegc: Interesting! Thanks for sharing @lucasaarcoverde !

marcelovicentegc: Indeed, when it comes to arrays, ternaries are preferentiable, but it occurs to me that this is exclusive to arrays.

Lenderson (approved):

Lenderson (approved):

anitavincent (approved):

packages/admin-ui-docs/docs/guidelines/components/search.mdx
```diff
@@ -119,59 +229,133 @@ function Example() {
...
+
+<PageSection
+  id={tabToc.codeReference[2].id}
+  title={tabToc.codeReference[2].value}
+  explanation="You can use all the props accepted by the `form` JSX element."
+>
+
+| Name     | Type                                   | Description                              | Required | Default |
+| -------- | -------------------------------------- | ---------------------------------------- | -------- | ------- |
+| onClear  | `() => void`                           | Clears the input value                   | ✅       | -       |
```
marcelovicentegc, line 278: Why is this requried?

packages/admin-ui/src/search/hooks/useSearchState.ts
```diff
@@ -0,0 +1,73 @@
...
+export interface GetInputPropsReturn {
+  value: string
+  onChange: ChangeEventHandler<HTMLInputElement>
+  onClear: () => void
+}
+
+export interface UseSearchStateParams {
+  initialValue?: string
+  defaultValue?: string
+  timeoutMs?: number
```
marcelovicentegc, line 62: What do you think of having this prop named as `timeout`, while also having a good jsDoc stating that it expects a value as milliseconds? I haven't seen `timeoutMs` before and properties or parameters named `timeout` are quite common across standard (and non-standard such as Axios) HTTP libraries (I know, I know, nothing to do with a UI library 😄  but why not?).

packages/admin-ui/src/search/search.style.ts
```diff
@@ -0,0 +1,64 @@
...
+    color: '$disabled',
+    '::placeholder': {
+      color: '$disabled',
+    },
+  },
+})
+
+export const form = style({
+  position: 'relative',
+  marginX: '1px',
```
marcelovicentegc, line 38:
```suggestion
  marginX: '0.06 rem',
```

marcelovicentegc: Pixel police passing by 👮‍♀️

marcelovicentegc (approved):
I've left a couple of suggestions but LGTM. Great work Lucas!

packages/admin-ui-docs/docs/guidelines/components/search.mdx
lucasaarcoverde: It isn't, the doc is wrong. Good catch!

PR #193, opened 2022-05-16
fix(text-input): a11y, spaces, and next terminology

matheusps:
- Fix form-control spaces to match Figma.
- Fix form-control message a11y.
- Fix the text-input click/focus behavior.
- Add the `next` terminology to the text-input to avoid confusion.

Admin UI should match Figma and be inclusive to every user.

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/form-control/form-control-message.tsx
+++ b/packages/admin-ui/src/form-control/form-control-message.tsx
@@ -22,7 +22,7 @@ export function FormControlMessage(props: FormControlMessageProps) {
         </Text>
       ) : null}
       {hasError ? (
-        <Text variant="detail" tone="critical">
+        <Text variant="detail" tone="critical" role="alert">
           {errorText}
         </Text>
       ) : null}
--- a/packages/admin-ui/src/form-control/form-control.tsx
+++ b/packages/admin-ui/src/form-control/form-control.tsx
@@ -10,7 +10,7 @@ export function FormControl(props: FormGroupOptions) {
 
   return (
     <FormControlProvider {...remainingProps}>
-      <Stack space="$l">{children}</Stack>
+      <Stack space="$m">{children}</Stack>
     </FormControlProvider>
   )
 }
--- a/packages/admin-ui/src/index.ts
+++ b/packages/admin-ui/src/index.ts
@@ -3,8 +3,10 @@ export * from '@vtex/admin-ui-react'
 export * from '@vtex/admin-ui-hooks'
 export * from '@vtex/admin-ui-util'
 
-export * from './center'
+export * from './next'
 export * from './experimental'
+
+export * from './center'
 export * from './inline'
 export * from './bleed'
 export * from './button'
@@ -12,7 +14,6 @@ export * from './stack'
 export * from './radio'
 export * from './switch'
 export * from './checkbox'
-export * from './text-input'
 
 export * from './deprecated-set'
 
--- a/packages/admin-ui/src/text-input/text-input.tsx
+++ b/packages/admin-ui/src/text-input/text-input.tsx
@@ -1,6 +1,6 @@
 import type { ReactNode, ComponentPropsWithoutRef, Ref } from 'react'
-import React, { forwardRef } from 'react'
-import { useId } from '@vtex/admin-ui-hooks'
+import React, { useRef, forwardRef } from 'react'
+import { useId, useForkRef } from '@vtex/admin-ui-hooks'
 
 import { TextInputContainer } from './text-input-container'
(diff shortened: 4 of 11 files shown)


packages/admin-ui/src/form-control/form-control-message.tsx
```diff
@@ -22,7 +22,7 @@ export function FormControlMessage(props: FormControlMessageProps) {
         </Text>
       ) : null}
       {hasError ? (
-        <Text variant="detail" tone="critical">
+        <Text variant="detail" tone="critical" role="alert">
```
matheusps, line 25: Critical messages [must have the `alert` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role).

packages/admin-ui/src/form-control/form-control.tsx
```diff
@@ -10,7 +10,7 @@ export function FormControl(props: FormGroupOptions) {
 
   return (
     <FormControlProvider {...remainingProps}>
-      <Stack space="$l">{children}</Stack>
+      <Stack space="$m">{children}</Stack>
```
matheusps, line 13: Figma has the `$m` token, and not `$l`.

packages/admin-ui/src/index.ts
```diff
@@ -3,16 +3,17 @@ export * from '@vtex/admin-ui-react'
 export * from '@vtex/admin-ui-hooks'
 export * from '@vtex/admin-ui-util'
 
-export * from './center'
+export * from './next'
```
matheusps, line 6: The same pattern of experimental components. The main difference between those is => that the next ones are ready to ship but lack documentation.

packages/admin-ui/src/text-input/text-input.tsx
```diff
@@ -26,16 +26,24 @@ export const TextInput = forwardRef(
     } = props
 
     const id = useId(defaultId)
+    const innerRef = useRef<HTMLInputElement>(null)
+
+    const focus = () => {
```
matheusps, line 31: Focus on the input if the container is clicked. This will avoid confusion while clicking on the prefix and suffix.

anitavincent (approved):

packages/admin-ui/src/form-control/form-control-message.tsx
kevinch: Is that linked in the docs?

packages/admin-ui/src/text-input/text-input.tsx
```diff
@@ -1,6 +1,6 @@
 import type { ReactNode, ComponentPropsWithoutRef, Ref } from 'react'
-import React, { forwardRef } from 'react'
-import { useId } from '@vtex/admin-ui-hooks'
+import React, { useRef, forwardRef } from 'react'
+import { useId, useForkRef } from '@vtex/admin-ui-hooks'
```
kevinch, line 3: Curious on the naming here `useForkRef`.

packages/admin-ui/src/form-control/form-control-message.tsx
lucasaarcoverde: good catch!

kevinch (approved):

packages/admin-ui/src/index.ts
lucasaarcoverde: I think we should use the same pattern (experimental) for both. The difference would be that these components will be shipped to stable faster

packages/admin-ui/src/form-control/form-control-message.tsx
matheusps: @kevinch, we can document it on the a11y section once the `form-control` documentation is available.

packages/admin-ui/src/text-input/text-input.tsx
Lenderson: It is similar to [this one](https://react-hooks.org/docs/useForkRef)

lucasaarcoverde (approved):

packages/admin-ui/src/index.ts
matheusps: Don't think so. They don't have an epic, and they don't have missing features. We also don't advocate for their usage.

packages/admin-ui/src/text-input/text-input.tsx
matheusps: The `outer ref` **forks** the `inner ref` => Which means that the `inner ref` forwards some functionality to the `outer ref`.

Lenderson (approved):

PR #201, opened 2022-05-18
Feat/nad 976 number field

Lenderson:
Create the component `NumberInput` which will replace `NumericStepper`

#### Types of changes
- [x] Requires change to documentation, which has been updated accordingly


packages/admin-ui/src/number-input/number-input.tsx
```diff
@@ -0,0 +1,117 @@
...
+  label?: ReactNode
+  value: number | string
+  step: number
+  min: number
+  max: number
+  optional?: boolean
+  error?: boolean
+  errorText?: ReactNode
+  helpText?: ReactNode
+  onChange: (value: number | string) => void
```
Lenderson, line 113: I am wondering if this "custom" onChange make sense, because the input value will be change by typing and by clicking the increment and decrement buttons of this component 🧐

packages/admin-ui/src/number-input/number-input.style.ts
```diff
@@ -0,0 +1,32 @@
+import { style } from '@vtex/admin-ui-core'
+
+export const container = style({
+  ':not(:hover) > button': {
+    display: 'none',
+  },
+})
+
+export const input = style({
+  '::-webkit-inner-spin-button ': {
```
marcelovicentegc, line 10: I'm wondering whether we should really go with this pseudoelement as [it is marked as non-standard and not recommended for production](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-inner-spin-button). I might be missing something!

packages/admin-ui/src/number-input/number-input.tsx
marcelovicentegc: I guess it doens't really make a difference if we are only accepting number inputs (even if they are converted from strings). Should it be required though?

marcelovicentegc: I guess it doens't really make a difference if we are only accepting number inputs (even if they are converted from strings). Should it be required though?

packages/admin-ui/src/number-input/number-input.tsx
```diff
@@ -0,0 +1,117 @@
...
+type JSXInputProps = ComponentPropsWithoutRef<'input'>
+
+interface NumberInputOptions {
+  prefix?: ReactNode
+  suffix?: ReactNode
+  label?: ReactNode
+  value: number | string
+  step: number
+  min: number
+  max: number
```
marcelovicentegc, line 108: Required? It occurs to me that we could have a default value for `step` (1), `min` (-Infinity), and `max` (Infinity).

marcelovicentegc, line 108: Required? It occurs to me that we could have a default value for `step` (1), `min` (-Infinity), and `max` (Infinity).

packages/admin-ui/src/number-input/number-input.style.ts
Lenderson: We have to hide the native spin button and replace it for our custom spin button. In the worst case, the native spin button will continue to appear. At the moment, this is the way of "accessing" the webkit spin button style to override it.

packages/admin-ui/src/number-input/number-input.state.ts
```diff
@@ -0,0 +1,115 @@
...
+        currentValue,
+        nextValue: currentValue - step,
+        min,
+        max,
+      })
+
+      setCurrentValue(newValue)
+      onChange(newValue)
+    } else {
+      // When the currentValue is a empty string and you click to decrement
```
lucasaarcoverde, line 54: I think it would be better for the default value to be always 0, this way you don't need to make this condition of an empty string. What do you think? Maybe is worth talking to davis

packages/admin-ui/src/number-input/number-input.state.ts
```diff
@@ -0,0 +1,115 @@
...
+      const newValue = min >= -step ? min : -step
+
+      setCurrentValue(newValue)
+      onChange(newValue)
+    }
+  }
```
lucasaarcoverde, line 60: Just a suggestion and you can do the same for the increment
```suggestion
  const handleDecrement: React.MouseEventHandler<HTMLButtonElement> = (value: number | string) => {
    const currentValue = isNumber(value) ? value : 0
    const newValue = getDecrementedValue({
      currentValue,
      nextValue: currentValue - step,
      min,
      max,
    })

    setCurrentValue(newValue)
    onChange(newValue)
  }
```

packages/admin-ui/src/number-input/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  min <= value && value <= max
+
+export const getIncrementedValue = ({
+  currentValue,
+  nextValue,
+  min,
+  max,
+}: NumberValidationProps): number => {
+  if (isValidNumber(currentValue, min, max)) {
+    return nextValue <= max ? nextValue : max
```
lucasaarcoverde, line 17:
```suggestion
    return Math.min(max, nextValue)
```

packages/admin-ui/src/number-input/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const getDecrementedValue = ({
+  currentValue,
+  nextValue,
+  min,
+  max,
+}: NumberValidationProps): number => {
+  if (isValidNumber(currentValue, min, max)) {
+    return nextValue >= min ? nextValue : min
```
lucasaarcoverde, line 31:
```suggestion
    return Math.max(min, nextValue)
```

packages/admin-ui/src/number-input/number-input.state.ts
```diff
@@ -0,0 +1,115 @@
...
+      min,
+      max,
+      onChange: handleChange,
+      ...props,
+    }
+  }
+
+  const getDecrementButtonProps = (props: ButtonProps) => {
+    return {
+      'aria-label': 'Decrement',
```
lucasaarcoverde, line 76: important to have this localized

packages/admin-ui/src/number-input/number-input.style.ts
```diff
@@ -0,0 +1,34 @@
+import { style } from '@vtex/admin-ui-core'
+
+export const container = style({
+  ':not(:hover):not(:focus-within) > button': {
```
lucasaarcoverde, line 4: Nice!

packages/admin-ui/src/number-input/number-input.tsx
```diff
@@ -0,0 +1,137 @@
...
+      event
+    ) => {
+      event.preventDefault()
+    }
+
+    return (
+      <FormControl error={error} optional={optional}>
+        {label && <FormControlLabel htmlFor={id}>{label}</FormControlLabel>}
+        <InputContainer
+          onClick={handleClick}
```
lucasaarcoverde, line 78: I think this makes the button lose its focus when clicking using the keyboard

packages/admin-ui/src/number-input/number-input.state.ts
```diff
@@ -0,0 +1,115 @@
...
+  getDecrementedValue,
+} from './utils'
+
+export const useNumberInput = (props: NumberInputState) => {
+  const { initialValue, step, min, max, onChange } = props
+  const [currentValue, setCurrentValue] = useState(initialValue)
+
+  const handleChange: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+    const { value } = event.target
+    const newValue = isNumberAsString(value) ? Number(value) : value
```
lucasaarcoverde, line 16: I think you should add a validation here to check if the input is bigger or lower than the min and max values. I'm not sure if this is expected:

Captura de Tela 2022-05-24 às 16 31 02

lucasaarcoverde (commented):
Just few suggestions

packages/admin-ui/src/number-input/number-input.tsx
Lenderson: I am still thinking he best way to solve this, because it is important to keep the focus on the component when we click the button.

packages/admin-ui/src/number-input/number-input.state.ts
Lenderson: Using this suggestion it would be easier, but the behavior would be different from the native behavior of `input[type=number]` when there is no value. This is to avoid different behavior when it is used the button or the keyboard. What do you think about this?

lucasaarcoverde (approved):

marcelovicentegc (approved):
LGTM

PR #252, opened 2022-06-15
Reorg filter label

anitavincent:

Reduce some code repetition + separate disclosure from base filter structure, that will be helpful when implementing different states.

Check filters on built storybook

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/filters/filter-applied-list-label.tsx
+++ b/packages/admin-ui/src/filters/filter-applied-list-label.tsx
@@ -1,21 +1,17 @@
 import React from 'react'
 import { tag } from '@vtex/admin-ui-react'
-import type { UseFilterMultipleReturn } from '..'
-import type { AnyObject } from '../..'
+import type { AnyObject } from '..'
 
-import * as style from '../filter.style'
+import * as style from './filter.style'
 
-export const MultipleItemsLabel = (props: {
+export const AppliedListLabel = (props: {
   appliedItems: AnyObject[]
-  state: UseFilterMultipleReturn<any>
+  renderItem: (item: AnyObject) => string
 }) => {
-  const {
-    appliedItems,
-    state: { getOptionLabel },
-  } = props
+  const { appliedItems, renderItem } = props
 
   const separator = appliedItems.length > 1 ? ',' : ''
-  const firstOptionLabel = getOptionLabel(appliedItems[0] || {}) || ''
+  const firstOptionLabel = renderItem(appliedItems[0] || {}) || ''
 
   const firstSelectedItemLabel = `${firstOptionLabel}${separator}`
 
--- a/packages/admin-ui/src/filters/filter-base.tsx
+++ b/packages/admin-ui/src/filters/filter-base.tsx
@@ -5,7 +5,7 @@ import { Button } from '../button'
 import { FilterPopoverFooter, FilterPopover } from './filter-popover'
 
 import type { GenericFilterStateReturn } from './filter/filter.state'
-import { ComboboxList } from 'ariakit'
+import { ComboboxList } from 'ariakit/combobox'
 
 import { tag } from '@vtex/admin-ui-react'
 
--- a/packages/admin-ui/src/filters/filter-base.tsx
+++ b/packages/admin-ui/src/filters/filter-base.tsx
@@ -3,7 +3,6 @@ import React, { useRef } from 'react'
 
 import { Button } from '../button'
 import { FilterPopoverFooter, FilterPopover } from './filter-popover'
-import { FilterDisclosure } from './filter-disclosure'
 
 import type { GenericFilterStateReturn } from './filter/filter.state'
 import { ComboboxList } from 'ariakit'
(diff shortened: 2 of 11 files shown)


packages/admin-ui/src/filters/filter-base.tsx
```diff
@@ -3,7 +3,6 @@ import React, { useRef } from 'react'
 
 import { Button } from '../button'
 import { FilterPopoverFooter, FilterPopover } from './filter-popover'
-import { FilterDisclosure } from './filter-disclosure'
 
 import type { GenericFilterStateReturn } from './filter/filter.state'
 import { ComboboxList } from 'ariakit'
```
Lenderson, line 8: seria melhor deixar o import mais específico para evitar possíveis problemas no tree shaking
```suggestion
import { ComboboxList } from 'ariakit/combobox'
```

packages/admin-ui/src/filters/filter-disclosure.tsx
```diff
@@ -2,24 +2,37 @@ import type { ReactNode } from 'react'
 import React from 'react'
 import { Button } from '../button'
 import { IconCaretUp } from '@vtex/phosphor-icons'
-import type { MenuState } from 'ariakit'
 import { MenuButton } from 'ariakit'
```
Lenderson, line 5: semelhante a [esse](https://github.com/vtex/admin-ui/pull/252/files#r898401971)

packages/admin-ui/src/filters/filter-base.tsx
```diff
@@ -17,35 +16,26 @@ import { Stack } from '../stack'
...
-      <FilterDisclosure state={menu} id={disclosureId}>
-        {label}
-        {appliedValuesLabel}
-      </FilterDisclosure>
-
       <FilterPopover state={menu} id={popoverId}>
         <tag.span ref={optionsContainerRef} csx={style.scrollableContainer}>
           <Stack
-            as={ComboboxList}
+            as={ComboboxList as any}
```
Lenderson, line 38: Why not use the ComboboxList component with the Stack inside of it?

packages/admin-ui/src/filters/filter-disclosure.tsx
```diff
@@ -2,24 +2,37 @@ import type { ReactNode } from 'react'
...
 export const FilterDisclosure = (props: FilterDisclosureProps) => {
-  const { state, children, id } = props
+  const { state, children, appliedItems } = props
+
+  const disclosureId = state.baseId ? `${state.baseId}-disclosure` : undefined
 
   return (
-    <Button as={MenuButton as any} state={state} csx={style.disclosure} id={id}>
+    <Button
+      as={MenuButton as any}
```
Lenderson, line 19: Maybe you could use the MenuButton and just pass to it our button styles. What do you think? Make sense in this case?

packages/admin-ui/src/filters/filter-multiple/filter-multiple.tsx
```diff
@@ -22,31 +23,30 @@ export function FilterMultiple<T>(props: FilterMultipleProps<T>) {
...
-            aria-checked={undefined}
-            csx={{ marginRight: '$s' }}
-            label={getOptionLabel(item)}
-            aria-readonly="true"
-          />
-        </ComboboxItem>
-      ))}
-    </BaseFilter>
+    <>
+      <FilterDisclosure state={state} appliedItems={appliedItems}>
```
Lenderson, line 27: Ithink the filter-multiple with and without search are almost the same, the only difference is the search and the items array, maybe a boolean prop (`searchable`) would solve this duplication?

packages/admin-ui/src/filters/filter/filter.tsx
```diff
@@ -26,32 +26,35 @@ export function Filter<T>(props: FilterProps<T>) {
...
-            style={style.option}
-            id={`${baseId ?? ''}-item-${itemId}`}
-          >
-            <FilterRadio checked={itemId === currentSelectedId} />
-            {getOptionLabel(item)}
-          </ComboboxItem>
-        )
-      })}
-    </BaseFilter>
+    <>
```
Lenderson, line 29: similar to the [previous case](https://github.com/vtex/admin-ui/pull/252/files#r898432463)?

packages/admin-ui/src/filters/filter-applied-list-label.tsx
```diff
@@ -1,21 +1,17 @@
...
 import { tag } from '@vtex/admin-ui-react'
-import type { UseFilterMultipleReturn } from '..'
-import type { AnyObject } from '../..'
+import type { AnyObject } from '..'
 
-import * as style from '../filter.style'
+import * as style from './filter.style'
 
-export const MultipleItemsLabel = (props: {
+export const AppliedListLabel = (props: {
```
Lenderson, line 7: suggestion: AppliedItemsLabel

packages/admin-ui/src/filters/filter-base.tsx
anitavincent: Not sure if it would break something acessibility-wise. Considering the list should have specific props if it's a combobox-list.

Either way I want to think about a better way to use these components other than "as" but I won't do that on this pr

packages/admin-ui/src/filters/filter-applied-list-label.tsx
```diff
@@ -1,21 +1,17 @@
...
   appliedItems: AnyObject[]
-  state: UseFilterMultipleReturn<any>
+  renderItem: (item: AnyObject) => string
```
Lenderson, line 9: I would change to something as `renderLabel` since its return is a string

packages/admin-ui/src/filters/filter-base.tsx
Lenderson: ok 😉

packages/admin-ui/src/filters/filter-disclosure.tsx
anitavincent: Might make sense, I'm still thinking about a better way to handle these "as" and will come back to it and other ariakit topics in the future

packages/admin-ui/src/filters/filter-multiple/filter-multiple.tsx
anitavincent: I'm not sure about this one. Yeah, they are alike, but they also have a bunch of tiny different things (filter search has many more status too) and I don't love code with many ifs to avoid repetition. 

But still I might merge them later, if I think about a fluid way. In truth all 4 filters could technically be one single component, but it would be a mess

Lenderson (approved):

packages/admin-ui/src/filters/filter-multiple/filter-multiple.tsx
```diff
@@ -22,31 +23,30 @@ export function FilterMultiple<T>(props: FilterMultipleProps<T>) {
...
+            aria-selected={combobox.isSelected(item)}
+            key={getOptionId(item)}
+            onClick={() => combobox.onChange(item)}
+            style={style.option}
+            id={`${baseId ?? ''}-item-${getOptionId(item)}`}
+          >
+            <Checkbox
+              id={getOptionId(item)}
+              checked={combobox.isSelected(item)}
+              aria-checked={undefined}
```
marcelovicentegc, line 42: Out of curiosity: why is this always `undefined`?

marcelovicentegc (approved):
LGTM

packages/admin-ui/src/filters/filter-multiple/filter-multiple.tsx
anitavincent: There is another aria check indicator above on the tree, this one would be a duplicate on the same item

PR #265, opened 2022-06-22
Refactor/table cell style

Lenderson:
1. Set `background-color` when a row is selected 
2. Let `variable` as the default value of table density
3. Set image `height` (image resolver)
4. Create a text resolver to show a text with description inside a cell

Check the `Selectable` example in storybook

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/table/components/table-cell.tsx
+++ b/packages/admin-ui/src/table/components/table-cell.tsx
@@ -9,7 +9,7 @@ export const TableCell = createComponent<'td', CellOptions>((props) => {
     onClick,
     csx,
     role = 'cell',
-    density = 'compact',
+    density = 'variable',
     ...cellProps
   } = props
 
--- a/packages/admin-ui/src/table/components/table-cell.tsx
+++ b/packages/admin-ui/src/table/components/table-cell.tsx
@@ -9,7 +9,7 @@ export const TableCell = createComponent<'td', CellOptions>((props) => {
     onClick,
     csx,
     role = 'cell',
-    density = 'regular',
+    density = 'compact',
     ...cellProps
   } = props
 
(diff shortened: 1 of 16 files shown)


packages/admin-ui/src/table/components/table-cell.tsx
```diff
@@ -9,7 +9,7 @@ export const TableCell = createComponent<'td', CellOptions>((props) => {
     onClick,
     csx,
     role = 'cell',
-    density = 'regular',
+    density = 'variable',
```
Lenderson, line 12: **PS:** I'm just setting every one to have `variable` as the default density and in the next PR I will remove this prop and let its style inside the baseline style.

packages/admin-ui/src/table/resolvers/image.tsx
```diff
@@ -18,6 +20,12 @@ const defaultPreview: ImagePreview = {
...
 
 function getImageVariant(density: TableDensity): StyleObject {
+  const outlineStyle = style({
+    outlineColor: (theme) => get(theme, 'bg.primary', 'bg.primary'),
+    outlineWidth: '0.125rem',
+    outlineStyle: 'solid',
+  })
+
   return {
     regular: {
```
Lenderson, line 30: Similar to this [case](https://github.com/vtex/admin-ui/pull/265/files#r906314635). I will set the variable style as the base style of the image.

packages/admin-ui/src/table/styles/table-cell.styles.ts
```diff
@@ -18,9 +26,8 @@ export const variants = styleVariants({
       height: 48,
     },
     variable: {
```
Lenderson, line 28: In the next PR it will be inside the baseline style.

packages/admin-ui/src/table/components/table-cell.tsx
Lenderson: I want to avoid in this PR a lot changes related to the `density` prop

anitavincent (approved):

PR #267, opened 2022-06-23
NAD-1153 - Page Header component review

marcelovicentegc:

To update the current Page Header component following the component review guidelines.

[Design specs are available on this Figma file.](https://www.figma.com/file/sepBhrHViUHyrTnktkuHaW/Admin-UI-%E2%80%93-Components?node-id=5869%3A389994)

[Visit this storybook story ](https://page-header-review.surge.sh/?path=/story/admin-ui-review-page-page-header--with-tabs)and its documentation section (see preview). Make sure it matches the Figma specs.

#### Types of changes


packages/admin-ui/src/page/page-header.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+          return
+        }
+
+        setScrollOnTop(false)
+        lastScrollY = scrollY > threshold ? scrollY : threshold
+        ticking = false
+      }
+
+      const onScroll = () => {
```
Lenderson, line 78: Was this created to add a box-shadow when the user scrolls down the page? 🤔 If the answer is yes, i think it could be better to use the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

marcelovicentegc: Good catch! Yes, that's the expected behavior!

marcelovicentegc: I decided not to go with the Intersection Observer API because it is mainly used for complex scrolling behaviors, such as infinite scrolls and elements that appear when a certain amount of the page has been scrolled. Since this is such a simple behavior, it occurs to me that an event listener could be the best approach. Do you see any downsides with the event listener approach vs intersection observer? cc @Lenderson

marcelovicentegc: Thanks for the tip, Lenderson! I developed a mixed approach using the Intersection Observer API + Portals. Indeed, it's unusual to see this use case being handled by the IO API, but nonetheless it is the best approach, as you pointed 😉

packages/admin-ui-hooks/src/use-intersection-observer.ts
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+    const { current: currentObserver } = observer
+
+    // We need to disconnect in case the component using
+    // this hook changes the node that the observer is
+    // observing, otherwise we'll polute the DOM with
+    // unused observables. This ensures that the observer
+    // is acting only over the node we care about.
+    currentObserver.disconnect()
```
Lenderson, line 39: When the useEffect dependencies change the cleanup function is called, so every time the node dependency changes, before run the useEffect, the previous cleanup function is called. Every time the node changes the `line 43` is called.

packages/admin-ui-hooks/src/use-intersection-observer.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    node && currentObserver.observe(node)
+
+    return () => currentObserver.disconnect()
```
Lenderson, line 43:
```suggestion
    if (!node) {
    return;
  }

  observer.current = new IntersectionObserver(
    ([entry]) => {
      callback(entry);
    },
    {
      root,
      rootMargin,
      threshold,
    }
  );

  const { current: currentObserver } = observer;

  return () => currentObserver.disconnect();
```

packages/admin-ui-hooks/src/use-intersection-observer.ts
marcelovicentegc: https://github.com/vtex/admin-ui/pull/267#discussion_r913135653

Lenderson (approved):

packages/admin-ui/src/page/page-header-bottom.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export const PageHeaderBottom = createComponent<'div'>((props) => {
+  const { children, ...htmlProps } = props
+  const hasTab = useRef(false)
+
+  Children.forEach(children, (child) => {
+    if ((child as any)?.type === TabList) {
+      hasTab.current = true
+    }
+  })
```
lucasaarcoverde, line 36: There are other use case to it without a TabList in the bottom?

packages/admin-ui/src/page/page-header-actions.ts
```diff
@@ -0,0 +1,46 @@
...
+ *    </PageHeaderActions>
+ *  </PageHeaderTop>
+ * </PageHeader>
+ */
+export const PageHeaderActions = createComponent<'div'>((props) => {
+  const { children, ...htmlProps } = props
+
+  return useElement('div', {
+    baseStyle: style.pageHeaderActions,
+    children,
```
lucasaarcoverde, line 41: Avoid the definition of spacing in the children, it's a better practice to always define this in a parent container. In this case you can use the `Stack` component to do it
```suggestion
    children: (
      <Stack space="$l" direction="row">
        {children}
      </Stack>
    ),
```

packages/admin-ui/src/page/page.style.ts
```diff
@@ -0,0 +1,143 @@
...
+  justifyContent: 'flex-end',
+  '> div': {
+    display: 'flex',
+  },
+  '> button:not([aria-haspopup="menu"])': {
+    minWidth: '6.5rem',
+  },
+  '> * + *': {
+    marginLeft: '$l',
+  },
```
lucasaarcoverde, line 111:
```suggestion
```

packages/admin-ui/src/page/page.style.ts
```diff
@@ -0,0 +1,143 @@
...
+})
+
+export const pageHeaderActions = style({
+  display: 'flex',
+  alignItems: 'center',
+  flexWrap: 'wrap',
+  justifyContent: 'flex-end',
+  '> div': {
+    display: 'flex',
+  },
```
lucasaarcoverde, line 105: Is it necessary?

packages/admin-ui-hooks/src/use-intersection-observer.ts
```diff
@@ -0,0 +1,42 @@
...
+  // Defaults to the browser viewport
+  root = null,
+  rootMargin = undefined,
+  threshold = 0,
+  callback,
+}: IntersectionObserverInit & {
+  callback: (entry: IntersectionObserverEntry) => void
+}) {
+  const [node, setNode] = useState<HTMLDivElement>()
+  const observer = useRef<IntersectionObserver>()
```
lucasaarcoverde, line 16: Just to follow our code styleguide: [Avoid param destructuring](https://admin-ui.vercel.app/teams/developing/code-styleguide#avoid-param-destructuring) and [Avoid inline types](https://admin-ui.vercel.app/teams/developing/code-styleguide#avoid-inline-types)

```suggestion
export function useIntersectionObserver(props: UseIntersectionObserverProps) {
  const {
    // Defaults to the browser viewport
    root = null,
    rootMargin = undefined,
    threshold = 0,
    callback,
  } = props
  
  const [node, setNode] = useState<HTMLDivElement>()
  const observer = useRef<IntersectionObserver>()
```

packages/admin-ui-hooks/src/use-portal.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export type UsePortalOptions = {
+  /**
+   * Attach the portal to this node in the DOM
+   */
+  bindTo?: HTMLElement
+} & CustomEventHandlers
+
+export function usePortal({ bindTo }: UsePortalOptions = {}): any {
```
lucasaarcoverde, line 32: Same

packages/admin-ui/src/page/page-header-title.tsx
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * <PageHeader>
+ *  <PageHeaderTop>
+ *    <PageHeaderTitle>
+ *      Product #123
+ *      <Stack direction="row" space="$m">
+ *        <Tag label="Short text" size="large" />
+ *        <Tag label="Short text" size="large" />
+ *      </Stack>
+ *    </PageHeaderTitle>
```
lucasaarcoverde, line 24: I'm trying to figure out a better approach to it. One option is to do the `Stack` space behavior inside the `PageHeaderTitle` csx, [like this](https://codesandbox.io/s/reverent-dew-1kelbf?file=/src/Root.tsx).
The bad part is that this requires to pass the `Product #123` label inside a `<p>` tag

lucasaarcoverde: Other option would be to have a `tags` property in the `PageHeaderTitle`, something like this:
```jsx
<PageHeaderTitle tags={[{label, tone}, {label, tone}]}>
  Product #123
</PageHeaderTitle>
```

lucasaarcoverde: And the last option I can imagine is to separate the Tag from the title. It looks to me they are two different things and that's why it's being difficult to find a better solution

```
<PageHeader>
  <PageHeaderTop>
    <PageHeaderTitle>...</PageHeaderTitle>
    <PageHeaderTags>
       <Tag />
       <Tag />
     </PageHeaderTags>
     <PageHeaderActions>...</PageHeaderActions>
  </PageHeaderTop>
</PageHeader>
```

packages/admin-ui/src/page/page-header-title.tsx
```diff
@@ -0,0 +1,54 @@
...
+            bleedX
+            size="large"
+            bleedY
+            icon={<IconArrowLeft />}
+            onClick={onPopNavigation}
+            csx={style.popNavigationButton}
+          />
+        )}
+        <tag.div csx={style.pageHeaderTitle}>{children}</tag.div>
+      </>
```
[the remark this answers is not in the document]
marcelovicentegc: This doesn't work as expected since the Button's `bleedY` must be taken into account 😊

packages/admin-ui/src/page/page-header-bottom.ts
marcelovicentegc: Nope! This is the only scenario for tabs

packages/admin-ui-hooks/src/use-intersection-observer.ts
marcelovicentegc: Done 😉

packages/admin-ui-hooks/src/use-portal.tsx
marcelovicentegc: Done 😉

packages/admin-ui/src/page/page.style.ts
marcelovicentegc: Not really, good catch!

packages/admin-ui/src/page/page.style.ts
marcelovicentegc: Applied 😉

packages/admin-ui/src/page/page-header-actions.ts
marcelovicentegc: Applied 😉

packages/admin-ui/src/page/page-header-title.tsx
marcelovicentegc: I'm happy with the current approach, where we got a loose API. This gives us the flexibility to let this component evolve without us having to change its code. 

It occurs to me that with good documentation, our clients will figure out how to add tags on their own, with no need for magic under the hood or explicit wrappers (like PageHeaderTags)

marcelovicentegc: [I have used the second approach before](https://github.com/vtex/admin-ui/pull/267/commits/bd8e30eda8585f04f06aaed0b30aefd85390d7d6#diff-046718c6255d4ea5aea39a1e01e02196383a13731e54626c2124cfab408688b3R96), but it looks like this is an anti-pattern for our project, since we are applying compound components to most of our "complex" components

packages/admin-ui/src/page/page-header-button.ts
```diff
@@ -0,0 +1,37 @@
...
+ *      Product #123
+ *    </PageHeaderTitle>
+ *    <PageHeaderActions>
+ *      <PageHeaderButton>Create</PageHeaderButton>
+ *      <PageHeaderButton variant="critical">Delete</PageHeaderButton>
+ *    </PageHeaderActions>
+ *  </PageHeaderTop>
+ * </PageHeader>
+ */
+export const PageHeaderButton = createComponent<typeof Button, ButtonOptions>(
```
lucasaarcoverde, line 25: Avoid using `createComponent` with a component created using `createComponent` :inception:. This can lead to some types and behavior errors since we don't cover this scenario
Do something like this instead
```jsx
export const PageHeaderButton = forwardRef((props: ButtonProps, ref: Ref<HTMLButtonElement>) => {
  const { size="large", ...buttonProps } = props
  return <Button ref={ref} size={size} {...buttonProps} />
})
```

marcelovicentegc: Nice! Thanks for the tip @lucasaarcoverde

anitavincent (approved):

packages/admin-ui/src/page/page-header-button.ts
marcelovicentegc: Done!

lucasaarcoverde (approved):

Lenderson (approved):

PR #270, opened 2022-06-27
Reduce styles complexity

matheusps:

This PR makes the styles function less complex and more powerful by:
- Removing the splits definition that lacked power, since you could only spread the same value across two props.
- Creating the utils definition. It replaces splits by using a function (it can deal with more things ,with less overhead).
- Removing the default prop of the theme (unused). 

#### Types of changes


packages/admin-ui-core/src/styles.ts
```diff
@@ -87,21 +87,9 @@ export function styles(csxObject: StyleProp = {}, theme: any = defaultTheme) {
...
 
     if (!!value && typeof value === 'object') {
-      // handle default entries on rules
-      if (value.default) {
-        // handle object rules
-        cssObject[cssProperty] =
-          typeof value.default === 'object'
-            ? styles(value.default, theme)
-            : value.default
-      } else {
```
matheusps, line 97: This was dead code.

lucasaarcoverde (approved):

PR #308, opened 2022-07-12
feat(filters): enable csx on all composable filter components

anitavincent:
Use admin ui core functions and enable csx usage on all filter components.

Check the csx example under FiltersSearch on storybook (ps the searchbox doesnt change a lot because our search component doesnt)

#### Types of changes


packages/admin-ui/src/filters/filter-disclosure.tsx
```diff
@@ -1,13 +1,14 @@
...
-import { Button } from '../button'
+
 import { IconCaretUp } from '@vtex/phosphor-icons'
 import { MenuButton } from 'ariakit/menu'
 
 import * as style from './filter.style'
 import { AppliedItemsLabel } from './filter-applied-items-label'
 import type { UseFilterMultipleReturn } from './filter-multiple/filter-multiple.state'
 import type { UseFilterStateReturn } from './filter/filter.state'
+import { createComponent, useElement } from '..'
```
lucasaarcoverde, line 11:
```suggestion
import { createComponent, useElement } from '@vtex/admin-ui-react'
```

packages/admin-ui/src/filters/filter-listbox.tsx
```diff
@@ -1,43 +1,48 @@
...
+    const scrollHeight = optionsContainerRef?.current?.scrollHeight ?? 0
+    const containerHeight = optionsContainerRef?.current?.clientHeight ?? 0
+
+    useEffect(() => {
+      setIsScrollableLayout(scrollHeight > containerHeight)
+    }, [scrollHeight, containerHeight])
+
+    const ariakitcomboboxState = { ...combobox, matches: [] }
+
+    return useElement('span', {
```
Lenderson, line 28: I think this should be a `div` because it is a problem have a inline element `span` wrapping a block element `div`.

packages/admin-ui/src/filters/filter-listbox.tsx
```diff
@@ -1,43 +1,48 @@
...
+    const {
+      setIsScrollableLayout,
+      state: { combobox },
+    } = usePopoverContext()
+
+    const scrollHeight = optionsContainerRef?.current?.scrollHeight ?? 0
+    const containerHeight = optionsContainerRef?.current?.clientHeight ?? 0
+
+    useEffect(() => {
+      setIsScrollableLayout(scrollHeight > containerHeight)
```
Lenderson, line 23: 🤔 Is this useEffect trigged only on the mount step? I just say that, because refs do not trigger re-renders...

anitavincent: No, the component is updated for other reasons and grows or shrinks. Then the heights are updated and if they changed this is triggered, for checking if the component became scrollable

It does work on real time

anitavincent: Actually this pr broke it 😆  thanks for making me double check

anitavincent: The issue was that I was overwriting ref with undefined from the props. It's fixed and now it works

Lenderson (approved):

lucasaarcoverde (approved):
Just a change to the import

PR #309, opened 2022-07-12
refactor(table): update table display to 'grid'

Lenderson:
Update table style using 'CSS Grid'

Inspect and check the ARIA `role` attribute of each table element.

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/table/components/table-head.tsx
+++ b/packages/admin-ui/src/table/components/table-head.tsx
@@ -1,19 +1,20 @@
 import React, { cloneElement, Fragment } from 'react'
-import { createComponent, useElement, tag } from '@vtex/admin-ui-react'
+import { createComponent, useElement } from '@vtex/admin-ui-react'
 import { IconArrowDown, IconArrowUp } from '@vtex/phosphor-icons'
 
 import { useStateContext } from '../context'
 import { TableCell } from './table-cell'
 import * as styles from '../styles/table-head.styles'
+import { Box } from '../../box'
 
-export const TableHead = createComponent<'thead'>((props) => {
-  const { children, role = 'rowgroup', ...headProps } = props
+export const TableHead = createComponent<'div'>((props) => {
+  const { children, ...headProps } = props
   const state = useStateContext()
 
-  return useElement('thead', {
+  return useElement('div', {
     baseStyle: styles.baseline,
     ...headProps,
-    role,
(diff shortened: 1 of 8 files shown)


packages/admin-ui/src/table/components/table-head.tsx
```diff
@@ -1,19 +1,26 @@
...
+export const TableHead = createComponent<'div'>((props) => {
+  const { children, ...headProps } = props
   const state = useStateContext()
 
-  return useElement('thead', {
+  const ariaSortLabel = {
+    ASC: 'ascending',
+    DESC: 'descending',
+    DEFAULT: 'none',
+  } as any
```
Lenderson, line 18: [Aria sort values](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaSort#value)

marcelovicentegc: Noice

packages/admin-ui/src/table/styles/table-row.styles.ts
```diff
@@ -1,8 +1,16 @@
...
+}
+export const rowBaseline = ({ columns }: RowOptions) =>
+  style({
+    display: 'grid',
+    bg: '$primary',
+    textAlign: 'left',
+    borderBottom: '$neutral',
+    gridTemplateColumns: columns
+      .map((column) => (column.width ? `${column.width}px` : '1fr'))
+      .join(' '),
```
marcelovicentegc, line 15: I've seen this pattern pop up a couple of times now. Maybe we should create a utility function to handle this grid template columns case?

marcelovicentegc (approved):
LATM (looks awesome to me)

packages/admin-ui/src/table/styles/table-row.styles.ts
Lenderson: It still will change a little in the next PR's, but the gridTemplateColumns changes depending on the use case.

lucasaarcoverde: I think [this is the other case](https://github.com/vtex/admin-ui/blob/dev/packages/admin-ui/src/grid/grid.tsx#L48-L50) where you saw this, but they are different. I'm not sure it's a pattern 🤔

lucasaarcoverde (approved):

PR #340, opened 2022-07-22
Feat/unify filter states

anitavincent:
Simplify filter state management by abstracting combobox functionalities and unifying search and global status

All filter stories

#### Types of changes


packages/admin-ui/src/filters/use-filter-status.ts
```diff
@@ -0,0 +1,57 @@
+import { useState, useMemo } from 'react'
+
+import type { ComboboxState } from '../combobox/combobox.state'
+
+import type { AnyObject } from '@vtex/admin-ui-util'
+import type { FilterStatus } from './filter/filter.state'
+
+export function useFilterStatus<T extends AnyObject>(props: ComboboxState<T>) {
```
lucasaarcoverde, line 8: The only difference between combobox status/setStatus and the new one created here are the status ids (loading-error, search-error, etc.)? If I get it right, IMO it would be better to have the same ids here and only reuse the combobox state

packages/admin-ui/src/filters/filter-popover.tsx
```diff
@@ -12,11 +12,17 @@ import { FilterStatus } from './filter-status'
...
+    const { children, state, onRetry = () => {}, ...restProps } = props
 
     const [isScrollableLayout, setIsScrollableLayout] = useState(false)
     const contextState = { isScrollableLayout, setIsScrollableLayout, state }
 
+    const shouldDisplayChildren = !(
+      state.status === 'error' ||
+      state.status === 'loading' ||
+      state.status === 'empty'
+    )
```
lucasaarcoverde, line 24: Is it the same result?
```suggestion
    const shouldDisplayChildren = state.status === 'ready'
```

packages/admin-ui/src/filters/filter/filter.state.tsx
```diff
@@ -84,11 +99,26 @@ export interface FilterOption<T> {
   value?: T
 }
 
+export type FilterStatus =
```
lucasaarcoverde, line 102: What's the difference between error/error-search and loading/loading-search?

packages/admin-ui/src/filters/filter-popover.tsx
```diff
@@ -12,11 +12,17 @@ import { FilterStatus } from './filter-status'
...
 export const FilterPopover = createComponent<typeof Menu, FilterPopoverOptions>(
   (props) => {
-    const { children, state, ...restProps } = props
+    const { children, state, onRetry = () => {}, ...restProps } = props
 
     const [isScrollableLayout, setIsScrollableLayout] = useState(false)
     const contextState = { isScrollableLayout, setIsScrollableLayout, state }
 
+    const shouldDisplayChildren = !(
+      state.status === 'error' ||
```
Lenderson, line 21: **Suggestion**: Probably, it would be nice to have all those status as a enum, i think it would help to know the possible filter status to be used. What do you think? 🤔

packages/admin-ui/src/filters/filter-status.tsx
```diff
@@ -46,6 +53,41 @@ export function FilterStatus(props: FilterStatusProps) {
...
+
+// {status === 'error' && (
+//   <Box csx={{ padding: '$l' }}>
+//     <Text variant="title2">{formatMessage('errorTitle')}</Text>
+//     <Button variant="tertiary" onClick={onRetry} bleedX>
+//       {formatMessage('retry')}
+//     </Button>
+//     <FilterFooter csx={{ marginTop: '$l', padding: 0 }} />
+//   </Box>
+// )}
```
matheusps, line 93:
```suggestion
```

packages/admin-ui/src/filters/filter-popover.tsx
Lenderson: it seems to have others status, but I am not entirely sure 🤔

packages/admin-ui/src/filters/filter-status.tsx
```diff
@@ -17,12 +19,17 @@ export function FilterStatus(props: FilterStatusProps) {
   if (!status) return null
 
   // more filter status comming soon
-  if (status === 'ready' || status === 'loading' || status === 'empty')
+  if (
+    status === 'ready' ||
+    status === 'loading' ||
+    status === 'empty' ||
+    status === 'error'
+  )
```
Lenderson, line 27: Suggestion:
```suggestion
 if (["ready", "loading", "empty", "error"].includes(status)) {
  return null
}

```

packages/admin-ui/src/filters/filter-status.tsx
anitavincent: oops

packages/admin-ui/src/filters/filter-status.tsx
anitavincent: Sounds ok, but why?

packages/admin-ui/src/filters/filter/filter.state.tsx
anitavincent: Error with search has a different ui, same for loading

packages/admin-ui/src/filters/filter-popover.tsx
anitavincent: No, all the search status should still display children

packages/admin-ui/src/filters/use-filter-status.ts
anitavincent: No, there are two loadings, global loading and search loading. And so on

packages/admin-ui/src/filters/filter/filter.state.tsx
matheusps: Maybe there is a UI issue here. That's why its so complex.

anitavincent: I don't think the design is unreasonable... The search states "loading" and "error" are different than global loading and global error. The difference between the search states is that the popover keeps showing children (it keeps the searchbox and buttons rendered)

Maybe I implemented it in a confusing way? These are all the states:

```
export type FilterStatus =
  | 'error' // replaces popover with error
  | 'loading' // replaces popover with skeleton, for initial loading of any kind of filter
  | 'ready' 
  | 'empty' // totally empty filter, no options to choose. This replaces the popover with a message
  | 'not-found' // shows not found message, but still shows searchbox and everything else. Doesn't interrupt the user typing
  | 'search-error' // shows error message, but still shows searchbox and everything else
  | 'loading-search' // just shows loading on searchbox
  ```

File: packages/admin-ui/src/filters/filter-status.tsx (hunk 6 above)
Comment (Lenderson) [reply packages/admin-ui/src/filters/filter-status.tsx:27]: It is more a suggestion, but it is just a first possible change, because we can see some status being checking more than once... I was just wondering if it could make sense let those checking more explicit...as creating a function which represents those checking...

**Comment by anitavincent** [comment]
Now the filter status are 

```
export type FilterStatus =
  | 'error'
  | 'loading'
  | 'ready' 
  | 'empty'
  | 'not-found' 
```

error and search-error were unified. Search-loading is handled as a searchbox prop, since this status just affects the searchbox and loading is already a prop of our `<Search />`

File: packages/admin-ui/src/filters/use-filter-status.ts (hunk 1 above)
Comment (lucasaarcoverde) [reply packages/admin-ui/src/filters/use-filter-status.ts:8]: Is this hook still useful since we have the same status?

File: packages/admin-ui/src/filters/use-filter-status.ts [hunk 7]
```diff
@@ -0,0 +1,29 @@
+import { useState, useMemo } from 'react'
+
+import type { ComboboxState } from '../combobox/combobox.state'
+
+import type { AnyObject } from '@vtex/admin-ui-util'
+import type { FilterStatus } from './filter/filter.state'
```
lucasaarcoverde, line 6: This generates a circular dependency, it's better to export this type from `use-filter-status.ts` file

lucasaarcoverde (approved):
LGTM, just a small change

PR #350, opened 2022-07-27
feat(css-vars): add function to convert the theme into css variables

lucasaarcoverde:

Adds a function to convert the theme into CSS Variables

#### Types of changes

### DIFF
--- a/packages/admin-ui-core/src/createTheme.ts
+++ b/packages/admin-ui-core/src/createTheme.ts
@@ -306,7 +306,7 @@ export function generateVars<T>(node: T, theme = {}, ruleId = '', accKey = '') {
   const vars: Record<string, object> = {}
 
   const isRoot = !accKey
-  const initialTheme = !isRoot ? node : theme
+  const initialTheme = isRoot ? node : theme
 
   for (const key in node) {
     const rule = isRoot ? key : ruleId
--- a/packages/admin-ui-core/src/createTheme.ts
+++ b/packages/admin-ui-core/src/createTheme.ts
@@ -2,8 +2,8 @@ import { pick, omit, get } from '@vtex/admin-ui-util'
 
 const constants = {
   /**
-   * When converting tokens with those namespaces to CSS Variables the algorithm will look
-   * to the color values and replace all usage of color arbitrary values with its correspondent color variable.
+   * When converting tokens with those namespaces to CSS Variables, the algorithm will look
+   * to the color values and replace all usage of arbitrary values with their corresponding color variable.
    */
   colorNamespaces: ['bg', 'border', 'fg', 'shadows'],
   /**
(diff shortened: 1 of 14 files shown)


packages/admin-ui-core/src/createTheme.ts
```diff
@@ -1,6 +1,11 @@
 import { pick, omit, get } from '@vtex/admin-ui-util'
 
 const constants = {
+  /**
+   * When converting tokens with those namespaces to CSS Variables the algorithm will look
+   * to the color values and replace all usage of color arbitrary values with its correspondent color variable.
```
Lenderson, line 6: I think **by** it is better for this case
```suggestion
   * to the color values and replace all usage of color arbitrary values by its correspondent color variable.
```

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -235,3 +240,86 @@ export function objectToVars(obj: Record<string, any>, parent = '') {
...
+    if (result.includes(colorValue)) {
+      result = result.replaceAll(colorValue, toVarValue(`colors-${key}`))
+    }
+  })
+
+  return result
+}
+
+/**
+ * Parses an theme recursivelly to css variables, joining the paths
```
Lenderson, line 284:
```suggestion
 * Parses a theme recursively to css variables, joining the paths
```

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -235,3 +240,86 @@ export function objectToVars(obj: Record<string, any>, parent = '') {
...
+ *
+ * // returns:
+ * {
+ *   '--admin-ui-colors-blue': 'blue',
+ *   '--admin-ui-colors-yellow': 'yellow',
+ *   '--admin-ui-bg-primary': 'var(--admin-ui-colors-blue)',
+ *   '--admin-ui-bg-secondary': 'var(--admin-ui-colors-yellow)',
+ * }
+ */
+export function generateVars<T>(obj: T, theme = {}, ruleId = '', parent = '') {
```
Lenderson, line 305: **Suggestion**: Maybe changing the `parent` param name could be good, I was expecting this param to be a object, but it is the key name which represents the "parent object" or something like it.

Lenderson (approved):
LGTM, just a few suggestions.

packages/admin-ui-core/src/createTheme.ts
anitavincent: Replace with is ok too

anitavincent (approved):

packages/admin-ui-core/src/createTheme.ts
lucasaarcoverde: I've changed this a little bit and I think `with` fits better, can you take a look?

Lenderson (approved):

PR #352, opened 2022-07-28
Refactor/override theme

Lenderson:

1. Create a function to load a custom style configuration from a config file.
2. Create a function to be able to remove global styles if needed.

#### Types of changes

### DIFF
--- a/packages/admin-ui-core/src/createTheme.ts
+++ b/packages/admin-ui-core/src/createTheme.ts
@@ -1,4 +1,4 @@
-import { pick, omit, get } from '@vtex/admin-ui-util'
+import { pick, omit, get, merge } from '@vtex/admin-ui-util'
 
 const constants = {
   /**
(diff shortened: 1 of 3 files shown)


packages/admin-ui-core/src/createTheme.ts
```diff
@@ -115,6 +115,79 @@ function createRootStylesAsObject(cssVariables: CSSVariables) {
   }, {})
 }
 
+/**
+ * Return the custom theme which will override the initial theme.
+ * The custom theme will be loaded from the admin-ui.config.js file found in the project.
+ */
+
```
matheusps, line 122:
```suggestion
```

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -115,6 +115,79 @@ function createRootStylesAsObject(cssVariables: CSSVariables) {
...
+ * // returns:
+ * {
+ *  colors: {
+ *    background: 'red',
+ *    fg: 'red',
+ *    text: 'black',
+ *  }
+ * }
+ */
+
```
matheusps, line 167:
```suggestion
```

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -115,6 +115,79 @@ function createRootStylesAsObject(cssVariables: CSSVariables) {
...
+    console.log('There is no theme config file')
+  }
+
+  return customTheme
+}
+
+/**
+ * Merge deeply two theme objects, overriding the initial theme props by the other one props.
+ * @example
+ * mergeThemes
```
matheusps, line 143:
```suggestion
 * mergeThemes(
```

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -115,6 +115,79 @@ function createRootStylesAsObject(cssVariables: CSSVariables) {
...
+ * {
+ *  colors: {
+ *    background: 'red',
+ *    fg: 'red',
+ *    text: 'black',
+ *  }
+ * }
+ */
+
+export function mergeThemes(
```
matheusps, line 168: You could just use `merge` from `@vtex/admin-ui-util`.

Lenderson: ok, I will check it!!

Lenderson: It worked, but I had to consider when `global: {}`, a way of disabling global styles

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -115,6 +115,66 @@ function createRootStylesAsObject(cssVariables: CSSVariables) {
...
+ * {
+ *  global: {},
+ *  colors: {
+ *    background: 'blue',
+ *    fg: 'black',
+ *    text: 'black',
+ *  },
+ * }
+ */
+export function removeInitialGlobalStyles(
```
lucasaarcoverde, line 167: Is this function really necessary? It looks to me that this can be replaced by something like this in the script:
```
import { omit } from '@vtex/admin-ui-util'
...

const initialTheme = disableGlobalStyles ? omit(theme, ['global'] : theme
```

packages/admin-ui-core/src/createTheme.ts
```diff
@@ -115,6 +115,66 @@ function createRootStylesAsObject(cssVariables: CSSVariables) {
...
+  initialTheme: Record<string, any>,
+  initialGlobalStylesDisabled = false
+): Record<string, any> {
+  if (initialGlobalStylesDisabled) {
+    return { ...initialTheme, global: {} }
+  }
+
+  return initialTheme
+}
```
lucasaarcoverde, line 176: What do you think of changing to something like this? I'm suggesting this because the global style isn't removed always, so the function name is not accurate

```suggestion
export function resolveGlobal(
  initialTheme: Record<string, any>,
  isGlobalDisabled = false
): Record<string, any> {
  if (isGlobalDisabled) {
    return { ...initialTheme, global: {} }
  }

  return initialTheme
}
```

lucasaarcoverde (approved):
LGTM ✨

PR #356, opened 2022-08-01
Feat filter async fetch

anitavincent:
Allow user to set full search list after initial load

This enables async list fetching along with uncontrolled search behavior

Check filtersSearch/async for new behavior. Other searchable filters should work as before

#### Types of changes


packages/admin-ui/src/filters/__stories__/filters-searchable.stories.tsx
```diff
@@ -107,30 +107,34 @@ export function MultiSearch() {
...
-    api('', 5000).then((res: any[]) =>
+    setStatus('loading')
+    api('', 5000).then((res: any[]) => {
       setFullList(res.map((i) => ({ id: i.id, label: i.country })))
-    )
+      setStatus('ready')
+    })
   }, [])
 
+  useEffect(() => {
```
anitavincent, line 122: this use effect is only necessary because after you set status for the first time it becomes controlled forever (you can't unset it) so you don't get automatic updates. Maybe I should allow status undefined, or have a status 'auto' idk. On the other hand I don't know if it's nice to mix controlled and uncontrolled behavior.

Opinions welcome

packages/admin-ui/src/filters/filter/filter.state.tsx
```diff
@@ -11,9 +12,10 @@ import { useFilterStatus } from '../use-filter-status'
 export function useFilterState<T extends AnyObject>(
   props?: UseFilterStateProps<T>
 ): UseFilterStateReturn<T> {
-  const { fullList = [] } = props || {}
+  const { fullList: initialFullList = [] } = props || {}
```
lucasaarcoverde, line 15: fullList represents the initial matches? What's the difference between matches, setMatches and fullList, setFullList?

anitavincent: fullList => list that is searched by combobox internally
matches => current matches from the fulllist. If the user is controlling everything they can ignore fulllist and just set matches in general

setFullList => you can use it to set the searchable list, and then let the component perform the searches
setMatches => you are the one controlling the searches and so you only set the "search result" everytime

lucasaarcoverde (approved):
LGTM

matheusps (approved):

PR #397, opened 2022-09-09
feat: create prop to enable text description to be truncated (ellipsis)

Lenderson:
Enable text description to be truncated using ellipsis.

#### Screenshots or example usage
Check the Text resolver section (Code tab)

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/table/resolvers/text.tsx
+++ b/packages/admin-ui/src/table/resolvers/text.tsx
@@ -1,10 +1,11 @@
 import type { ReactNode } from 'react'
 import React from 'react'
 import invariant from 'tiny-invariant'
+import { style } from '@vtex/admin-ui-core'
 
 import { Skeleton } from '../../skeleton'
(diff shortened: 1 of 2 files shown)


packages/admin-ui/src/table/resolvers/text.tsx
```diff
@@ -47,6 +65,7 @@ export function textResolver<T>() {
 export type TextResolver<T> = {
   type: 'text'
   columnType?: 'name' | 'text'
+  ellipsis?: boolean
```
lucasaarcoverde, line 68: What do you think of something like that? Instead of being a boolean, IMO this can more scalable if we want to add another option of overflow in the future
```suggestion
  overflow?: 'ellipsis' | 'auto'
```

Lenderson: good suggestion ;)

lucasaarcoverde (approved):

PR #451, opened 2022-11-08
Custom collapsible

anitavincent:
Implement a collapsible from card

Check story for card and for customcollapsible

#### Types of changes


packages/admin-ui/src/custom-collapsibe/custom-collapsible.stories.tsx
```diff
@@ -0,0 +1,168 @@
...
+export default {
+  title: 'admin-ui/CollapsibleCard',
+  component: Card,
+} as Meta
+
+export const SingleCollapsible = () => {
+  const {
+    getToggleProps,
+    getCollapseProps,
+    visible: isExpanded,
```
mairatma, line 31: I noticed that within `useCollapsible` we're converting a variable named `isExpanded` to be returned as `visible`. But here we convert from `visible` to `isExpanded`. I feel like it'd be best to choose just one of these two as the default: either keep the original `isExpanded` name or let this one be just `visible`, as it arrives.

mairatma (commented):
Lgtm! Not approving just because this is still marked as Draft. What would be left here, just tests?

packages/admin-ui/src/custom-collapsibe/use-collapsible.ts
```diff
@@ -0,0 +1,47 @@
+import { useCallback } from 'react'
```
mairatma, line 1: Just not sure about where the best place for this hook file is. Maybe we should have a `hooks` folder for these cases, where we have hooks that are not attached to a specific admin-ui component? So this could be at `src/hooks/use-collapsible.tsx` and its story file could be moved to `src/hooks/use-collapsible.stories.tsx`.

packages/admin-ui/src/custom-collapsibe/custom-collapsible.stories.tsx
marcelovicentegc: I would go with `visible` since this syntax is already used in other components (example: Modal and Menu, probably other pop-up components)

packages/admin-ui/src/custom-collapsibe/use-collapsible.ts
marcelovicentegc: AFAIK we're kind of using this pattern already. For instance, on the menu component we have a `menu.state.ts` file that just exports a hook (useMenuState): https://github.com/vtex/admin-ui/blob/dev/packages/admin-ui/src/menu/menu.state.ts. Same for the tab component (`tab.state.ts`): https://github.com/vtex/admin-ui/blob/dev/packages/admin-ui/src/tab/tab.state.ts

I would vote for renaming this file to `collapsible.state.ts` to keep consitent.

mairatma: Ah, got it! Let's follow the existing pattern then

lucasaarcoverde: I agree. The file structure can be similar to this:
```
src
  |- collapsible
     |- collapsible.state.ts
     |- stories
     |- tests
     |- index.ts 
```

matheusps (approved):

lucasaarcoverde (approved):

PR #581, opened 2023-01-27
refactor(divider): change divider to new component model

anitavincent:
Change dividier to new component model

Performance and consistency 

yarn test, checking storybook

#### Types of changes


packages/admin-ui/src/components/Divider/divider.css.ts
```diff
@@ -0,0 +1,31 @@
...
+//       borderLeft: 0,
+//       height: 'auto',
+//     },
+//   },
+// })
+
+// export const baseline = style({
+//   border: '$neutral',
+//   margin: '$space-0',
+// })
```
lucasaarcoverde, line 19: Should this be removed?

packages/admin-ui/src/components/Divider/index.tsx
```diff
@@ -1,27 +1,32 @@
...
+        data-orientation={orientation}
+        {...restProps}
+      />
+    )
   }
 )
 
-export type DividerOptions = VariantProps<typeof style.variants>
+type DividerOrientation = 'vertical' | 'horizontal'
+interface DividerProps extends ComponentPropsWithoutRef<'hr'> {
```
lucasaarcoverde, line 28: You can let `ComponentPropsWithRef` instead, because it's forwarding the ref
```suggestion
interface DividerProps extends ComponentPropsWithRef<'hr'> {
```

lucasaarcoverde (approved):
LGTM, just two suggestions

packages/admin-ui/src/components/Divider/index.tsx
matheusps: It should be without a ref I guess...

matheusps (approved):
LGTM after the first suggestion

packages/admin-ui/src/components/Divider/index.tsx
lucasaarcoverde: The only difference is that when using `WithRef` the `ref` property will be shown when using the component, right?

PR #613, opened 2023-03-01
refactor(stack): migrate to the new component model

lucasaarcoverde:

As the title says.

🚨 BREAKING CHANGE 🚨: Stack no longer supports the csx prop, you must use className instead.

#### Types of changes


packages/admin-ui/src/stack/stack.tsx
```diff
@@ -0,0 +1,71 @@
...
+      ref={ref}
+      {...htmlProps}
+      data-direction={responsiveDirection}
+      data-fluid={responsiveFluid}
+      className={cx(stackTheme(responsiveAlign, responsiveSpace), className)}
+    />
+  )
+})
+
+Stack.displayName = 'Stack'
```
marcelovicentegc, line 48: # displayName

The displayName string is used in debugging messages. Usually, you don’t need to set it explicitly because it’s inferred from the name of the function or class that defines the component. You might want to set it explicitly if you want to display a different name for debugging purposes or when you create a higher-order component, see [Wrap the Display Name for Easy Debugging](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) for details.

https://reactjs.org/docs/react-component.html#displayname
```suggestion
```

marcelovicentegc (approved):

packages/admin-ui/src/stack/stack.tsx
lucasaarcoverde: When creating the component using the `createComponent` function this was necessary, I forgot to remove it. Good catch!

PR #619, opened 2023-03-03
refactor(button): migrate to new component model

anitavincent:
#### Types of changes


packages/admin-ui/src/button/button.tsx
```diff
@@ -1,110 +1,96 @@
...
+  variant = 'primary',
+  iconPosition: defaultIconPosition = 'start',
+  loading = false,
+  disabled = false,
+  bleedY = false,
+  bleedX = false,
+  className = '',
+  children,
+  ...props
+}: ButtonProps) => {
```
lucasaarcoverde, line 32:
```suggestion
export const useButton = (props: ButtonProps) => {

const {
  icon,
  size = 'normal',
  variant = 'primary',
  iconPosition: defaultIconPosition = 'start',
  loading = false,
  disabled = false,
  bleedY = false,
  bleedX = false,
  className = '',
  children,
  ...htmlProps
} = props
```

lucasaarcoverde: Should we keep the `useComponent` hook pattern in the new component model?

packages/admin-ui/src/combobox/combobox.style.ts
```diff
@@ -1,4 +1,5 @@
-import { style, styleVariants } from '@vtex/admin-ui-core'
+import { csx, cx, dataAttr, style, styleVariants } from '@vtex/admin-ui-core'
+import { action as actionColorScheme } from '../button/button.css'
```
lucasaarcoverde, line 2: It looks to me that the "actionSchema" is used in several places, does it make sense to be something from the button itself, or should it be a utility?

IMO we can let it be attached to the button for now, but once our design language gets more mature it will probably make sense to turn it into something bigger. Saying that I would export it as `actionSchema`, what do you think?
```suggestion
import { actionSchema } from '../button/button.css'
```

packages/admin-ui/src/filters/filter.style.tsx
```diff
@@ -127,3 +136,8 @@ export const visibilitySelectorItemVariants = styleVariants({
     },
   },
 })
+
+export const visibilitySelectorMenuButtonTheme = cx(
+  buttonTheme,
+  csx({ display: 'flex', justifyContent: 'center', alignItems: 'center' })
+)
```
lucasaarcoverde, line 143: I would duplicate only the necessary style here since they are not the same. What do you think?
```suggestion
export const visibilitySelectorMenuButtonTheme = csx({
  {duplicated styles}
  display: 'flex',
  justifyContent: 'center',
  alignItems: 'center'
})
```

anitavincent: This case is kind of tricky because this is really pretty much a tertiary button. It uses the color scheme, sizing, everything. But it has an icon that changes direction and no inner containers. So the display had to be flex and centered.

The actual component is using data attributes too ```className={style.visibilitySelectorMenuButtonTheme}
            data-variant="neutralTertiary"
            data-size="normal"```
            
I'm not sure what is the best approach here. This component is supposed to be ariakits menu button, but with our button styling

packages/admin-ui/src/button/button.tsx
anitavincent: We didn't discuss that, I don't think it was a rule on the former model either. It could easily be changed and the hooks content integrated on the component. One advantage I see for keeping the hook for the button specifically is that it could enable someone to use our button stylings with a different element than "button", since we won't have the "as" prop anymore. Someone could use this hook to generate the props and plug them on a different element

packages/admin-ui/src/filters/filter.style.tsx
lucasaarcoverde: IMO the best approach is to duplicate. In the end, it will be something like that, right?

```ts
export const menuButtonTheme = csx({
  text: '$action1',
  border: 'none',
  borderRadius: '$base',
  cursor: 'pointer',
  position: 'relative',
  display: 'flex',
  justifyContent: 'center',
  alignItems: 'center'
  [dataAttr('variant', 'neutralTertiary')]: action({
    tone: 'neutral',
    variant: 'tertiary',
  }),
  [dataAttr('size', 'normal')]: {
    padding: '$space-2 $space-3',
    height: '2.25rem',
  },
```

lucasaarcoverde: This can be made in another PR since the main focus here is the button and not the filter

packages/admin-ui/src/button/button.tsx
lucasaarcoverde: >  I don't think it was a rule on the former model either.

That's true, maybe we can define it on the new component model. IMO it can be useful as well

packages/admin-ui/src/filters/filter.style.tsx
anitavincent: Yes something like that, but the data attributes wouldnt even be necessary anymore, just apply padding, height, etc on  root

packages/admin-ui/src/combobox/combobox.style.ts
anitavincent: Agreed. I think the schema should probably be reviewed too

lucasaarcoverde (approved):
It looks like the `useButton` is imported in some place and the build is breaking after the removal of this hook. You just fix it and LGTM

lucasaarcoverde (commented):
Just a suggestion

packages/admin-ui/src/modal/modal-button.tsx
```diff
@@ -1,26 +1,20 @@
...
 
-    const buttonProps: AnyObject = {
-      ref,
-      className: cx(csx(baseStyle), className),
-      ...buttonHtmlProps,
-    }
+    const { children, ...dismissProps } = useDialogDismiss(restProps)
+    const buttonProps = dismissModal
+      ? { ...dismissProps, children: restProps.children }
+      : restProps
```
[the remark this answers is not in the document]
anitavincent: Doesn't work. I tested it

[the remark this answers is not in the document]
anitavincent: The thing is, it does overwrite. The thing is when there's no children prop from the user, the dismissProps ends up with a children prop that is just the icon "X". So for example in a button without children but with icon (case of the close button in the modal) the button was displaying 2 "x"s. The one from icon, and the one from children generated by dismissprops

This problem was not here before because the buttonprops was generated before so it overwrote "children" with it's own children

packages/admin-ui/src/combobox/combobox.style.ts
marcelovicentegc: Do we have an issue to keep track of this future change?

PR #650, opened 2023-03-28
feat(filters): migrate to the new component model

lucasaarcoverde:

As the title says.
Breaking change: The filter and all its composites don't accept the csx property, you must use className instead

#### Types of changes


packages/admin-ui/src/filters/filter-listbox.tsx
```diff
@@ -1,61 +1,55 @@
...
+
+  useEffect(() => {
+    const isScrollable = scrollHeight > containerHeight
+
+    if (isScrollableLayout !== isScrollable) {
+      setIsScrollableLayout(isScrollable)
+    }
+  }, [scrollHeight, containerHeight])
+
+  const ariakitcomboboxState = { ...combobox, matches: [], visible: true }
```
marcelovicentegc, line 34:
```suggestion
  const ariakitComboboxState = { ...combobox, matches: [], visible: true }
```

packages/admin-ui/src/filters/filter-listbox.tsx
```diff
@@ -1,61 +1,55 @@
...
+  return (
+    <div
+      ref={useForkRef(optionsContainerRef, ref as any)}
+      className={cx(filterListboxTheme, className)}
+      data-status={status}
+      id={id}
+      {...htmlProps}
+    >
+      <ComboboxList
+        state={ariakitcomboboxState}
```
marcelovicentegc, line 45:
```suggestion
  const ariakitComboboxState = { ...combobox, matches: [], visible: true }

  return (
    <div
      ref={useForkRef(optionsContainerRef, ref as any)}
      className={cx(filterListboxTheme, className)}
      data-status={status}
      id={id}
      {...htmlProps}
    >
      <ComboboxList
        state={ariakitComboboxState}
```

packages/admin-ui/src/filters/filter-multiple/filter-option-checkbox.tsx
```diff
@@ -1,16 +1,66 @@
...
+      className={cx(filterControlInputTheme, className)}
+      onClick={(e) => {
+        combobox.onChange(item)
+        onClick?.(e)
+      }}
+      id={id}
+      aria-selected={combobox.isSelected(item)}
+      {...htmlProps}
+    >
+      {children || (
```
marcelovicentegc, line 49: Do we make sure somewhere that this isn't an empty object (truthy value)?

marcelovicentegc (approved):

anitavincent (approved):

packages/admin-ui/src/filters/filter-footer.tsx
```diff
@@ -1,32 +1,44 @@
...
+
 import { Button } from '../button'
 import { useMessageFormatter } from '../i18n'
 import { messages } from './filter.i18n'
 import { usePopoverContext } from './filter-popover-context'
 import { Stack } from '../stack'
+import { filterFooterTheme } from './filter.css'
 
-import * as style from './filter.style'
+export const FilterFooter = forwardRef(function FilterFooter(
```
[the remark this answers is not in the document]
lucasaarcoverde: It's just a personal preference to avoid anonymous functions (`(props) => {}`)

packages/admin-ui/src/filters/filter-multiple/filter-option-checkbox.tsx
lucasaarcoverde: I don't think we need to care about this case because it's the react children, so if there is an empty object is because the user did it, so it's an expected behavior

PR #652, opened 2023-03-29
feat(segment): migrate to the new component model

lucasaarcoverde:

As the title says.
Breaking change: The segment don't accept the csx property, you must use className instead

This component doesn't have a story, you can check the `DateField` story to test it.

#### Types of changes


packages/admin-ui/src/segment/segment.tsx
```diff
@@ -0,0 +1,255 @@
...
+          }
+
+          break
+        }
+
+        default:
+          e.preventDefault()
+          e.stopPropagation()
+          if ((isNumeric(e.key) || /^[ap]$/.test(e.key)) && !isReadOnly) {
+            // TODO: fix typing experience
```
marcelovicentegc, line 155: Worth creating an issue?

matheusps (approved):

packages/admin-ui/src/segment/segment.tsx
lucasaarcoverde: I don't think so, this is part of an experimental component and we'll revisit it anyways in the future

Issue #703, opened 2023-06-13
Number Input: down/up buttons submit form


lafray:
# Bug Report

- [x] I've read and understood the [Contributing guidelines](https://admin-ui.vercel.app/contributing/onboarding) and have done my best effort to follow them.
- [x] I've searched for any related issues and avoided creating a duplicate issue.

## What happened

With the new version of admin-ui (0.136.0) the buttons voted to submit by default, because of this the up/down button of the number imput is submitting the form

## What I expected to happen

Clicking the up/down button should not submit the form

## Reproducing

- open https://codesandbox.io/s/form-validation-number-input-q9ps9m?file=/src/App.tsx
- click in up or down buttons

davicostalf:
@lafray This issue was already solved, right?

lafray:
Yes, in 0.136.1 version

PR #704, opened 2023-06-13
Fix button type to button instance of submit

lafray:

Fix button type to button in page header go back and number input up/down
related in https://github.com/vtex/admin-ui/issues/703

Not submit form in popnavigation of header
Not submit form in up/down of number input

#### Types of changes

### DIFF
--- a/packages/admin-ui/src/page/page-header-title.tsx
+++ b/packages/admin-ui/src/page/page-header-title.tsx
@@ -46,6 +46,7 @@ export const PageHeaderTitle = forwardRef(function PageHeaderTitle(
       {onPopNavigation && (
         <Button
           aria-label="Back"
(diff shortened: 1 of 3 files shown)


packages/admin-ui/src/page/page-header-title.tsx
```diff
@@ -46,6 +46,7 @@ export const PageHeaderTitle = forwardRef(function PageHeaderTitle(
       {onPopNavigation && (
         <Button
```
matheusps, line 47: You could solve this on the `Button` itself, right? Define the property `type`, with a default value of `button`. It would solve a lot of use-cases, I suppose. What do you think?

matheusps (approved):

PR #711, opened 2023-06-16
Fix form control spacing

davicostalf:
I wanted to change the space only on the Radio Group and Checkbox group components, but I ended up breaking Input components. I think "Form Control" should only be used by Radio Group and Checkbox Group components, and there should be a "Form Input" component for the inputs. Radio and Checkbox are controls and the other fields are inputs.

But since it might take sometime to make this change, and the inputs are broken, I'm reverting my previous commit in this PR.

image


davicostalf:
Created an issue to evolve Form Control: https://github.com/vtex/admin-ui/issues/712

matheusps (approved):

PR #1653, opened 2024-06-25
fix(collection-view): change to actions only appear if onError or onEmpty are passed

### LINKED ISSUE #1652: [collection] Render button to action in  error and empty status even if you have not received an action
### Summary

In the current implementation of CollectionView, the action button for the status empty and error status is allways rendered, even if you have not received an action for the corresponding status. 
The functions for 'Error and notEmpty are not mandatory, therefore, if the status is one of these values ​​and its function was not passed, a button that does nothing is rendered.
I also think it is important to consider that there are scenarios in which there are no actions available. An example is the Meli integration promotions center, which currently has nothing that can be done as we only support integrations created by mel

### Expected behavior

It is expected that if you did not receive a function for the action of that status, the button will not be rendered in that status

### Current behavior

_No response_

### Reproducible Example

_No response_

### Suggested solution

_No response_

### Additional context

_No response_

### Package

@vtex/shoreline

### Packages version

_No response_

### Browser

_No response_

### Package Manager

pnpm

lafray:
fix #1652

## Summary

Explaned in the [issue](https://github.com/vtex/shoreline/issues/1652)

## Examples

image

### DIFF
--- a/packages/shoreline/src/components/collection/collection-view.tsx
+++ b/packages/shoreline/src/components/collection/collection-view.tsx
@@ -57,19 +57,14 @@ export const CollectionView = forwardRef<HTMLDivElement, CollectionViewProps>(
       )
     }
 
+    const handleActionByStatus: Record<string, (() => void) | undefined> = {
+      error: onError,
+      empty: onEmpty,
+    }
+    const handleAction = handleActionByStatus[status]
     const heading = getMessage(`${status}-heading`)
     const description = getMessage(`${status}-description`)
-    const action = getMessage(`${status}-action`)
-
-    function handleAction() {
-      switch (status) {
-        case 'error':
-          return onError?.()
-
-        case 'empty':
-          return onEmpty?.()
-      }
-    }
+    const action = handleAction && getMessage(`${status}-action`)
 
     const actionVariant = status === 'empty' ? 'primary' : 'secondary'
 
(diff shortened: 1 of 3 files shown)


packages/shoreline/src/components/collection/collection-view.tsx
```diff
@@ -57,19 +57,14 @@ export const CollectionView = forwardRef<HTMLDivElement, CollectionViewProps>(
       )
     }
 
+    const handleActionByStatus: Record<string, (() => void) | undefined> = {
```
matheusps, line 60: What if I pass the function "onError" and the status is `empty`? In this case empty state will have an action, right?

matheusps (changes requested):

packages/shoreline/src/components/collection/collection-view.tsx
lafray: No. The action button will only appear if the corresponding action function has been passed.

If it passes onError and not onEmpty, it renders the action button only in status=error
If it passes onError and onEmpty, renders action button in both statuses
If you pass onEmpty and not onError, it renders the action button only in status=empty
If neither onError nor onEmpty passes, the button will not be rendered in any state

matheusps (approved):

packages/shoreline/src/components/collection/collection-view.tsx
matheusps: ok

PR #1780, opened 2024-07-31
feat(icon): add IconChatCircle

anitavincent:
## Summary

add IconChatCircle


matheusps (changes requested):
you need to export it on the `src/index.ts` as well

matheusps (approved):

PR #1856, opened 2024-08-22
fix(toast-stack): pass custom ariaProps to the Toast component

### LINKED ISSUE #1536: ToastStack: `ariaProps` are not passed down
# Bug Report

- [ ] I've read and understood the [Contributing guidelines](https://admin-ui.vercel.app/contributing/onboarding) and have done my best effort to follow them.
- [X] I've searched for any related issues and avoided creating a duplicate issue.

## What happened

When using `<ToastStack>` component, the `toastOptions.ariaProps` option is not passed down to the HTML as the example in [react-hot-toast docs](https://react-hot-toast.com/docs/use-toaster#:~:text=id%7D%20%7B-,...toast.ariaProps,-%7D%3E).

## What I expected to happen

I'm using `ariaProps` to add the `role: 'status'` property to the Toast HTML, but it doesn't work. I also suggest adding a default role when the user doesn't use `ariaProps`, to improve accessibility.

```ts
<ToastStack
  toastOptions={{
    ariaProps: {
      role: 'status',
      'aria-live': 'polite',
    },
  }}
/>
```

## Reproducing

- shoreline-components version in your `package.json`: `@vtex/shoreline-components@0.64.4`
- Live link with error: not today 😛

viniciuslagedo:
## Summary

Resolves #1536 

## Examples

```typescript
// Renders the Toast component with role equal alert and aria-live equal assertive
<ToastStack toastOptions={
        {
          ariaProps: {
            role: 'alert',
            "aria-live": 'assertive'
          }
        }
      } />
```

### DIFF
--- a/packages/shoreline/src/components/toast/toast-stack.tsx
+++ b/packages/shoreline/src/components/toast/toast-stack.tsx
@@ -53,6 +53,7 @@ export function ToastStack(props: ToastStackProps) {
               id={t.id}
               loading={t.type === 'loading'}
               variant={(t as any).variant as any}
+              ariaProps={toastOptions?.ariaProps || t.ariaProps}
             >
               {(t as any).message}
             </Toast>
--- a/packages/shoreline/src/components/toast/toast-stack.tsx
+++ b/packages/shoreline/src/components/toast/toast-stack.tsx
@@ -2,8 +2,8 @@ import type { CSSProperties } from 'react'
 import type { DefaultToastOptions } from 'react-hot-toast/headless'
 import { useToaster } from 'react-hot-toast/headless'
 
-import { ToastAppear } from './toast-appear'
 import { Toast } from './toast'
+import { ToastAppear } from './toast-appear'
 
 /**
  * Toasts can appear at any time to provide instant feedback on actions. They are usually temporary, but can also require the user to dismiss.
@@ -53,7 +53,8 @@ export function ToastStack(props: ToastStackProps) {
               id={t.id}
               loading={t.type === 'loading'}
               variant={(t as any).variant as any}
-              ariaProps={toastOptions?.ariaProps || t.ariaProps}
+              {...t.ariaProps}
+              {...toastOptions?.ariaProps}
             >
               {(t as any).message}
             </Toast>
--- a/packages/shoreline/src/components/toast/toast.tsx
+++ b/packages/shoreline/src/components/toast/toast.tsx
@@ -1,5 +1,6 @@
 import type { MouseEventHandler, ReactNode } from 'react'
 import { Children, isValidElement } from 'react'
+import type { ToastOptions } from 'react-hot-toast'
 import { toast as hotToast } from 'react-hot-toast/headless'
 import {
   IconCheckCircleFill,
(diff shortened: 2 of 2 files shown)


packages/shoreline/src/components/toast/toast-stack.tsx
```diff
@@ -53,6 +53,7 @@ export function ToastStack(props: ToastStackProps) {
               id={t.id}
               loading={t.type === 'loading'}
               variant={(t as any).variant as any}
+              ariaProps={toastOptions?.ariaProps || t.ariaProps}
```
lucasaarcoverde, line 56: Did you try this approach? I think it has the same behavior since we pass the `toastOptions` to the `useToaster` hook
```suggestion
              {...t.ariaProps}
```

packages/shoreline/src/components/toast/toast.tsx
```diff
@@ -23,12 +24,24 @@ import { Button } from '../button'
...
  */
 export function Toast(props: ToastProps) {
-  const { id, variant = 'informational', children, loading, onDismiss } = props
+  const {
+    id,
+    variant = 'informational',
+    children,
+    loading,
+    onDismiss,
+    ariaProps,
```
lucasaarcoverde, line 33:
```suggestion
    ...restProps,
```

packages/shoreline/src/components/toast/toast.tsx
```diff
@@ -23,12 +24,24 @@ import { Button } from '../button'
...
 
   const icon = loading ? <Spinner /> : getIcon(variant)
 
   return (
-    <div data-sl-toast data-loading={loading} data-variant={variant}>
+    <div
+      data-sl-toast
+      data-loading={loading}
+      data-variant={variant}
+      {...ariaProps}
```
lucasaarcoverde, line 43:
```suggestion
      {...restProps}
```

packages/shoreline/src/components/toast/toast-stack.tsx
lucasaarcoverde: got it, in this case, I would recommend something like that, what do you think?
```suggestion
              {...t.ariaProps}
              {...toastOptions?.ariaProps}
```

viniciuslagedo: Sounds good ma friend! 🚀

viniciuslagedo: To make that change, we must set Toast props to `role` and `aria-live` and remove the property `ariaProps` itself.

viniciuslagedo: That's intentional? Just to clarify the suggestion

lucasaarcoverde: it's not necessary to specify those properties on the interface because we are spreading the props to a `div` element it should work as expected. What you can do to be more typesafe is to extend the ToastProps from `React.ComponentPropsWithoutRef<'div'>`

viniciuslagedo: I don't know if I understand this right, but I made a new commit.

[Check this out!](https://github.com/vtex/shoreline/pull/1856/commits/2f7da58aaba6508e7e4487a61736124450fc9ea4)

lucasaarcoverde: What I mean is that the `ToastOptions['ariaProps']` are all native props such as `role` and `aria-live` so is not necessary to create a `ariaProps: ToastOptions['ariaProps']` to pass those props to the Toast component, you can just spread them as you did in this new solution

lucasaarcoverde (approved):

PR #1866, opened 2024-08-22
docs(contributiong/code/component model): add information about internationalization

### LINKED ISSUE #1777: Document how to override default messages on components
## Summary

There are some Shoreline components that when you are building, you will need to add a translated default label so that the user doesn't need to add it every time the component is used. Those translations are important for consistency across applications, shoreline usage, and a better development experience. You can find an example of it on the heading title and the button labels from the `ConfirmationModal`:

Screenshot 2024-07-31 at 10 58 45

We already have a solution on how to add it which is documented [here](https://shoreline.vtex.com/guides/code/component-model#internationalization), but it is missing information on how to allow overriding those default labels.

## Solution

You must add the messages override step on the [Internationalization documentation](https://shoreline.vtex.com/guides/code/component-model#internationalization).

The solution is adding a `messages` property to the component with the correct typing. You can use the [CollectionView](https://github.com/vtex/shoreline/blob/main/packages/shoreline/src/components/collection/collection-view.tsx#L129-L154) as a reference to document this.

lafray:
## Summary

Add info about the pattern that must be followed to enable override of component messages

fix #1777

## Examples

image

### DIFF
--- a/packages/docs/pages/guides/code/component-model.mdx
+++ b/packages/docs/pages/guides/code/component-model.mdx
@@ -82,10 +82,10 @@ function Search(props) {
 
 ### Provide overwriting capability
 
-In most cases, it is necessary to allow the user of the component to override the internal messages.
+Some cases require the user to override the internal messages.
 
-You do this by adding a property called `messages` and passing as argument to `useMessage`.
-The `messages` type must contain the same keys used in your files from `messages` folder.
+You do this by adding a property called `messages` and passing it as an argument to the `useMessage` hook.
+The `messages` type must contain the same keys defined in your files from the `messages` folder.
 
 ```tsx
 import { createMessageHook } from '@vtex/shoreline'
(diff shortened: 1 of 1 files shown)


packages/docs/pages/guides/code/component-model.mdx
```diff
@@ -76,9 +76,44 @@ const useMessage = createMessageHook(messages)
...
+  return <input placeholder={getMessage('search-label')} {...props}/>
 }
 ```
 
+### Provide overwriting capability
+
+In most cases, it is necessary to allow the user of the component to override the internal messages.
+
+You do this by adding a property called `messages` and passing as argument to `useMessage`.
+The `messages` type must contain the same keys used in your files from `messages` folder.
```
lucasaarcoverde, line 88: What do you think?
```suggestion
Some cases require the user to override the internal messages. 

You do this by adding a property called `messages` and passing it as an argument to the `useMessage` hook.
The `messages` type must contain the same keys defined in your files from the `messages` folder.
```

lafray: Done!

lucasaarcoverde (approved):

PR #1875, opened 2024-08-23
docs(all): fix issue 1598

kevinch:
## Summary

Fixes https://github.com/vtex/shoreline/issues/1598


kevinch:
There is a Vercel deployment issue but I can't see which one from here.

lucasaarcoverde:
> There is a Vercel deployment issue but I can't see which one from here.

The issue is with the permission on Vercel. Sorry, we haven't solved it yet! I'll run it locally to test it when reviewing

packages/docs/utils/all-components-list.ts
```diff
@@ -31,7 +33,7 @@ export const allComponentsList = [
   'text',
   'textarea',
   'time-input',
-  'toast-stack',
+  'toast',
```
lucasaarcoverde, line 36: You need to change the `/public/assets/all-toast-stack.png` filename to `toast.png` otherwise it will break the UI

lucasaarcoverde (changes requested):

packages/docs/utils/all-components-list.ts
kevinch: Updated.

lucasaarcoverde (approved):
LGTM, thanks for the contribution! ✨

PR #1887, opened 2024-08-26
fix(toast-function): pass dismiss method from lib

### LINKED ISSUE #1605: ToastStack: no toast cleanup
### Summary

Shoreline uses a static-global-variable approach to keep toast state, but does not provide a way to clean current toasts.

The `react-hot-toast` library does provide a way to [dismiss toasts programatically](https://react-hot-toast.com/docs/toast#:~:text=Dismiss%20toast%20programmatically), but Shoreline does not reexport it.

The lack of this important feature currently makes us unable to cleanup any test code that uses the `ToastStack` component. Example:

```tsx
// If it was possible, cleanup code would be here
// beforeEach(() => {
//   toast.remove()
// })

it('Shows toast after button is pressed', async () => {
  const user = userEvent.setup()
  render(
    <ToastStack>
      <button onClick={() => toast.success('Success!')}>Show toast</button>
    </ToastStack>
  )
  
  const button = screen.getByRole('button', { name: 'Show toast' })
  await user.click(button)
  await waitFor(() => {
    expect(screen.queryByText('Success!')).toBeVisible()
  })
})

it('Does not show toast from previous test LOL', async () => {
  render(
    <ToastStack /> // ToastStack component uses a static global variable to store toast state.
    // Therefore, this renders the "Success!" from the previous test
    // (which is bad and there's no way to run cleanup code)
  )
  
  // This test fails because "Success" is actually visible
  expect(screen.queryByText('Success!')).not.toBeInTheDocument()
})
```

### Expected behavior

_No response_

### Current behavior

_No response_

### Reproducible Example

_No response_

### Suggested solution

_No response_

### Additional context

_No response_

### Package

@vtex/shoreline

### Packages version

_No response_

### Browser

_No response_

### Package Manager

yarn

viniciuslagedo:
## Summary

Resolves #1605 

## Examples

```typescript
// To cleanup all toasts
<Button
  onClick={() => {
    toast.dismiss()
  }}
>
  Cleanup all
</Button>

// To cleanup specific toast
<Button
  onClick={() => {
    const id = toast.loading('Loading...')
    setTimeout(() => toast.dismiss(id), 2000)
  }}
>
  Cleanup loading toast after 2 seconds!
</Button>
```

### DIFF
--- a/packages/shoreline/src/components/toast/stories/examples.stories.tsx
+++ b/packages/shoreline/src/components/toast/stories/examples.stories.tsx
@@ -1,6 +1,6 @@
-import { ToastStack, toast } from '../index'
-import { Stack } from '../../stack'
 import { Button } from '../../button'
+import { Stack } from '../../stack'
+import { ToastStack, toast } from '../index'
 
 export default {
   title: 'components/toast',
@@ -56,6 +56,21 @@ export function ToastFunction() {
         }}
       >
         Promise
+      </Button>
+      <Button
+        onClick={() => {
+          toast.dismiss()
+        }}
+      >
+        Cleanup all
+      </Button>
+      <Button
+        onClick={() => {
+          const id = toast.loading('Loading...')
+          setTimeout(() => toast.dismiss(id), 2000)
+        }}
+      >
+        Cleanup loading toast after 2 seconds!
       </Button>
     </Stack>
   )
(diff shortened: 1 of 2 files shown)


lucasaarcoverde (approved):
LGTM

PR #1961, opened 2024-09-19
docs: adds callout in the how to contribute page

beatrizmilhomem:


lucasaarcoverde (approved):

davicostalf:
@beatrizmilhomem Is this the best place for the warning? It seemed kinda lost to me there. Maybe right in the beginning of the documentation. Also, instead of "Do not" we usually write "Don't".

PR #1965, opened 2024-09-23
fix(drawer): fix layout shift

matheusps:
## Summary

This PR fixes the layout shift caused by scrollbars. In fact, this was a problem in the vaul library. I fixed the problems by updating to the latest version. [More details here](https://github.com/emilkowalski/vaul/releases/tag/v0.9.4)


packages/shoreline/src/themes/sunrise/components/drawer.css
```diff
@@ -22,7 +22,7 @@
     --size: var(--size-medium);
   }
 
-  &[vaul-drawer-direction="right"] {
+  &[data-vaul-drawer-direction='right'] {
```
vaporwavie, line 25: @matheusps shouldn't this be double quoted just like the other ones from the sheet?

davicostalf (approved):
LGTM

PR #1982, opened 2024-09-27
fix(filter-value): show children text content instead of value

### LINKED ISSUE #1505: FilterItem: `value` prop
# Support or Feedback

- [X] I've searched for any related issues/questions on Github before submitting this request.

## Context

Shoreline's component `FilterItem` receives a `value` and a `children` prop (it works like the `<select>`/`<option>` HTML tags).

image

The `children` text is presented at each of the `Filter` popover options. But after an option is "selected" then "applied", the `value` prop is what is actually shown by the `Filter` button (in the following image, look at the `Roles: 123` text):

image

## Suggestion

My suggestion is to change this behavior, to show the `children` prop text instead:

image

This would make the `<Filter>`/`<FilterOption>` behavior work more like HTML's `<select>`/`<option>` tags, making the `children` prop related to the visuals and the `value` prop related to React's controlled state. This would allow us to use `ID` strings in the `value` prop without harming UI visuals - and without constantly converting `ID`s<=>`string`s within controlled filters.

## Additional Suggestion

If the previous suggestion is accepted, I would also suggest accepting the type `number` in the `value` prop, so there's no need for constant conversion from/to numeric `ID`s (`id.toString()`/`Number.parseInt(value, 10)`).``

viniciuslagedo:
## Summary
Partially resolves #1505 

Now, the **inner text** from the children of the selected item is used on the selected value. Pay attention to the inner text expression here because it is not precisely the children property. So, no HTML tags will be rendered. I see that as a quick solution to not change all the state logic created.

### Simple
For this example, when the item with value `rj` is selected, the filter will show 'Cidade: Rio de Janeiro'

```typescript
const [city, setCity] = useState<string>('')

<Filter label="Cidade" value={city} setValue={setCity}>
  <FilterItem value="rj">Rio de Janeiro</FilterItem>
  <FilterItem value="sp">São Paulo</FilterItem>
  <FilterItem value="mg">Minas Gerais</FilterItem>
 </Filter>
```

### Multiple
For this example, when the items with values `rj` and `mg` are selected in that order, the filter will show 'Cidade: Rio de Janeiro, +1'

```typescript
const [cities, setCities] = useState<string>([])

<Filter label="Cidade" value={cities} setValue={setCities}>
  <FilterItem value="rj">Rio de Janeiro</FilterItem>
  <FilterItem value="sp">São Paulo</FilterItem>
  <FilterItem value="mg">Minas Gerais</FilterItem>
 </Filter>
```

### DIFF
--- a/packages/shoreline/src/components/filter/filter-value.tsx
+++ b/packages/shoreline/src/components/filter/filter-value.tsx
@@ -1,5 +1,6 @@
+import { useSelectContext } from '@ariakit/react'
 import type { ComponentPropsWithoutRef } from 'react'
-import { forwardRef } from 'react'
+import { forwardRef, useCallback } from 'react'
 import { useFilterContext } from './filter-context'
 
 const valueSeparator = ': '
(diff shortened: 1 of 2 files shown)


packages/shoreline/src/components/filter/filter-value.tsx
```diff
@@ -13,14 +14,25 @@ export const FilterValue = forwardRef<HTMLSpanElement, FilterValueProps>(
     const filter = useFilterContext()
     const value = filter?.useState('value')
 
-    const isArray = Array.isArray(value)
-    const isLongArray = isArray && value.length > 1
+    const valueIsArray = Array.isArray(value)
+    const valueIsMultiSelected = valueIsArray && value.length > 1
+
+    const selectContext = useSelectContext()
+    const items = selectContext?.useState('items')
```
rafarubim, line 21: Seems like this [was deprecated](https://ariakit.org/reference/use-select-store#usestate) in favor of [useStoreState](https://ariakit.org/reference/use-store-state)

viniciuslagedo: Basically the useStoreState does not exist yet in that version

rafarubim (approved):

PR #1989, opened 2024-10-04
docs: adds a page about accessing the Figma libraries

beatrizmilhomem:


davicostalf:
The videos are great! Two things I've noticed:
1. In the VTEX users section you state "Any other user from VTEX has a viewer access to any file.". I think its, worth noting that even designers dont' have edit access, and that the IT workflow should used for non-designers to request access to Figma. Once in Figma, the Shoreline team is open and anyone has viewer access to it.
1. The callout for contribution where it's currently placed seems to be related only to external users. Maybe a global position at the top would make more sense.

PR #2023, opened 2024-10-22
fix(shoreline): avoid infinite loops on the dev script

matheusps:
## Summary

The dev script was not providing a great DX. It was watching for all the project, causing infinite loops on the /dist folder. For that I:

- Changed the chokidar watcher for the parcel watcher.
- Removed useless deps on the Shoreline package.


vaporwavie (approved):
LGTM, valid migration to `@parcel/watcher`.

PR #2067, opened 2025-01-28
fix(filter-clear): disable clear button when value is empty array

Sophmrs:
## Summary
Solves this issue: https://github.com/vtex/shoreline/issues/2064

Now multi select filter's clear button renders as disabled correctly when the value is an empty array, not only with empty strings (or other falsy values). The behaviour is now the same for single item (radio) or multiple item (checkbox) filters.

## Examples

This behavior can be observed in the `Filter` `Show` story:
### This multiselect filter:
```jsx
<Filter label="Multiple status" defaultValue={[]}>
  <FilterItem value="Stable">Stable</FilterItem>
  <FilterItem value="Experimental">Experimental</FilterItem>
  <FilterItem value="Deprecated">Deprecated</FilterItem>
</Filter>
```
renders the clear button as disabled when it has no value and nothing is selected
image
and still renders it enabled when an item is selected
image
or if it has an item already applied
image

### While this single selection filter:
```jsx
<Filter label="Single status">
  <FilterItem value="Stable">Stable</FilterItem>
  <FilterItem value="Experimental">Experimental</FilterItem>
  <FilterItem value="Deprecated">Deprecated</FilterItem>
</Filter>
```
still renders as it did before, as expected, here disabled with no value
image
here enabled with an item selected
image
and here enabled with an item already applied
image

### DIFF
--- a/packages/shoreline/src/components/filter/filter-clear.tsx
+++ b/packages/shoreline/src/components/filter/filter-clear.tsx
@@ -26,12 +26,19 @@ function useFilterClear() {
   const selectValue = select?.useState('value') ?? ''
   const filterValue = filter?.useState('value') ?? ''
 
+  const isSelectValueEmpty = Array.isArray(selectValue)
+    ? selectValue.length === 0
+    : !selectValue
+  const isFilterValueEmpty = Array.isArray(filterValue)
+    ? filterValue.length === 0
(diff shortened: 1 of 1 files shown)


vaporwavie (approved):
nice work! thanks for the contribution.

PR #2079, opened 2025-02-11
chore: add component into dkportal ADMT-654

marcelovicentegc:
This PR adds this component into our software catalog :hugs: 

> Also removes a folder that was hanging there with no use.


vaporwavie (approved):
Good stuff!

PR #2137, opened 2026-04-06
Bugbot

FabianoVFranca:


matheusps:
I'm rejecting this PR because we already have coderabbit on the repository right now. Makes sense?

PR #2148, opened 2026-08-12
fix: document body not reachable when drawer is open

lucasaarcoverde:
## Summary

- Unlocks `document.body` `pointer-events` while a controlled `DrawerProvider` is open, so portaled overlays (e.g. `ConfirmationModal`) remain clickable.
- Works around Vaul `1.1.2` not forwarding `modal={false}` from `Drawer.Root` ([vaul#580](https://github.com/emilkowalski/vaul/pull/580); [radix#3141](https://github.com/radix-ui/primitives/issues/3141)).
- Adds a Storybook play test for opening/confirming/canceling a confirmation modal from inside the drawer.

## Examples

Do not change the component API.

https://github.com/user-attachments/assets/7b1fc441-42da-423a-b304-8ed74e9550bc

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved interaction with non-modal drawers so the page remains responsive while a drawer is open.
  - Fixed pointer interaction issues when displaying confirmation modals inside drawers.
  - Ensured page interaction is restored reliably when drawers close.

- **Tests**
  - Added coverage for opening, confirming, and canceling modals displayed within drawers.
  - Added verification that pointer interactions are restored when drawers open and close.

### DIFF
--- a/packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
+++ b/packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
@@ -5,6 +5,7 @@ import {
   getByText,
   getByTestId,
   expect,
+  waitFor,
 } from '@storybook/test'
 import { ConfirmationModal } from '../../../confirmation-modal'
 import { Button } from '../../../button'
@@ -18,30 +19,35 @@ import { DrawerContent } from '../../drawer-content'
 
 export default {
   title: 'components/drawer/tests',
-  play: async () => {
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement)
     // DrawerPopover portals outside #storybook-root
     const body = within(document.body)
+
+    await userEvent.click(canvas.getByRole('button', { name: 'Open drawer' }), {
+      delay: 200,
+    })
+
+    await waitFor(() =>
+      expect(document.body.style.pointerEvents).not.toBe('none')
+    )
 
     const openModalButton = body.getByRole('button', {
       name: 'Open confirmation',
     })
+    const modal = getByTestId(document.body, 'confirmation-modal')
 
     await userEvent.click(openModalButton, { delay: 200 })
-
-    const modal = getByTestId(document.body, 'confirmation-modal')
     await expect(modal).toBeVisible()
 
-    const confirmButton = getByText(modal as HTMLElement, 'Confirm')
+    const confirmButton = getByText(modal, 'Confirm')
     await userEvent.click(confirmButton, { delay: 200 })
-    await expect(modal).not.toBeVisible()
+    await waitFor(() => expect(modal).not.toBeVisible())
 
     await userEvent.click(openModalButton, { delay: 200 })
-    const openModal = getByTestId(document.body, 'confirmation-modal')
-    const cancelButton = getByText(openModal as HTMLElement, 'Cancel')
+    const cancelButton = getByText(modal, 'Cancel')
     await userEvent.click(cancelButton, { delay: 200 })
-    await expect(
-      getByTestId(document.body, 'confirmation-modal')
-    ).not.toBeVisible()
+    await waitFor(() => expect(modal).not.toBeVisible())
   },
   parameters: {
     chromatic: { disableSnapshot: true },
@@ -49,6 +55,7 @@ export default {
 }
 
 export function ModalInsideDrawer() {
+  const [drawerOpen, setDrawerOpen] = useState(false)
   const [modalOpen, setModalOpen] = useState(false)
 
   const handleClose = () => {
@@ -56,25 +63,30 @@ export function ModalInsideDrawer() {
   }
 
   return (
-    <DrawerProvider open>
-      <DrawerPopover size="small">
-        <DrawerHeader>
-          <DrawerHeading>Drawer Heading</DrawerHeading>
-          <DrawerDismiss />
-        </DrawerHeader>
-        <DrawerContent>
-          <Button onClick={() => setModalOpen(true)}>Open confirmation</Button>
-          <ConfirmationModal
-            data-testid="confirmation-modal"
-            open={modalOpen}
-            onClose={handleClose}
-            onConfirm={handleClose}
-            onCancel={handleClose}
-          >
-            <Text variant="body">This is a confirmation modal</Text>
-          </ConfirmationModal>
-        </DrawerContent>
-      </DrawerPopover>
-    </DrawerProvider>
+    <>
+      <Button onClick={() => setDrawerOpen(true)}>Open drawer</Button>
+      <DrawerProvider open={drawerOpen} onOpenChange={setDrawerOpen}>
+        <DrawerPopover size="small">
+          <DrawerHeader>
+            <DrawerHeading>Drawer Heading</DrawerHeading>
+            <DrawerDismiss />
+          </DrawerHeader>
+          <DrawerContent>
+            <Button onClick={() => setModalOpen(true)}>
+              Open confirmation
+            </Button>
+            <ConfirmationModal
+              data-testid="confirmation-modal"
+              open={modalOpen}
+              onClose={handleClose}
+              onConfirm={handleClose}
+              onCancel={handleClose}
+            >
+              <Text variant="body">This is a confirmation modal</Text>
+            </ConfirmationModal>
+          </DrawerContent>
+        </DrawerPopover>
+      </DrawerProvider>
+    </>
   )
 }
--- a/packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
+++ b/packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
@@ -1,0 +1,80 @@
+import { useState } from 'react'
+import {
+  userEvent,
+  within,
+  getByText,
+  getByTestId,
+  expect,
+} from '@storybook/test'
+import { ConfirmationModal } from '../../../confirmation-modal'
+import { Button } from '../../../button'
+import { Text } from '../../../text'
+import { DrawerProvider } from '../../drawer-provider'
+import { DrawerPopover } from '../../drawer-popover'
+import { DrawerHeader } from '../../drawer-header'
+import { DrawerHeading } from '../../drawer-heading'
+import { DrawerDismiss } from '../../drawer-dismiss'
+import { DrawerContent } from '../../drawer-content'
+
+export default {
+  title: 'components/drawer/tests',
+  play: async () => {
+    // DrawerPopover portals outside #storybook-root
+    const body = within(document.body)
+
+    const openModalButton = body.getByRole('button', {
+      name: 'Open confirmation',
+    })
+
+    await userEvent.click(openModalButton, { delay: 200 })
+
+    const modal = getByTestId(document.body, 'confirmation-modal')
+    await expect(modal).toBeVisible()
+
+    const confirmButton = getByText(modal as HTMLElement, 'Confirm')
+    await userEvent.click(confirmButton, { delay: 200 })
+    await expect(modal).not.toBeVisible()
+
+    await userEvent.click(openModalButton, { delay: 200 })
+    const openModal = getByTestId(document.body, 'confirmation-modal')
+    const cancelButton = getByText(openModal as HTMLElement, 'Cancel')
+    await userEvent.click(cancelButton, { delay: 200 })
+    await expect(
+      getByTestId(document.body, 'confirmation-modal')
+    ).not.toBeVisible()
+  },
+  parameters: {
+    chromatic: { disableSnapshot: true },
+  },
+}
+
+export function ModalInsideDrawer() {
+  const [modalOpen, setModalOpen] = useState(false)
+
+  const handleClose = () => {
+    setModalOpen(false)
+  }
+
+  return (
+    <DrawerProvider open>
+      <DrawerPopover size="small">
+        <DrawerHeader>
+          <DrawerHeading>Drawer Heading</DrawerHeading>
+          <DrawerDismiss />
+        </DrawerHeader>
+        <DrawerContent>
+          <Button onClick={() => setModalOpen(true)}>Open confirmation</Button>
+          <ConfirmationModal
+            data-testid="confirmation-modal"
+            open={modalOpen}
+            onClose={handleClose}
+            onConfirm={handleClose}
+            onCancel={handleClose}
+          >
+            <Text variant="body">This is a confirmation modal</Text>
+          </ConfirmationModal>
+        </DrawerContent>
+      </DrawerPopover>
+    </DrawerProvider>
+  )
+}
--- a/packages/shoreline/src/components/drawer/tests/use-non-modal-drawer-unlock.test.tsx
+++ b/packages/shoreline/src/components/drawer/tests/use-non-modal-drawer-unlock.test.tsx
@@ -57,7 +57,7 @@ describe('useNonModalDrawerUnlock', () => {
     fireEvent.click(getByRole('button', { name: 'Open drawer' }))
 
     await waitFor(() => {
-      expect(document.body.style.pointerEvents).not.toBe('none')
+      expect(document.body.style.pointerEvents).toBe('auto')
     })
   })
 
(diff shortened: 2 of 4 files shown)


packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export function ModalInsideDrawer() {
+  const [modalOpen, setModalOpen] = useState(false)
+
+  const handleClose = () => {
+    setModalOpen(false)
+  }
+
+  return (
+    <DrawerProvider open>
```
matheusps: **This story opens the drawer at mount — the one configuration vaul already handles — so this test is green on `main` without the fix.**

`open` is set at mount here, and vaul's own workaround covers exactly that case: `Root` runs

```js
useEffect(() => {
  if (!modal) requestAnimationFrame(() => { document.body.style.pointerEvents = 'auto' })
}, [modal])   // vaul/dist/index.mjs:1331-1339
```

once at mount. Radix's `DismissableLayer` writes `'none'` from a passive effect during the same commit, so the rAF lands *after* it and wins.

I ran this story's exact shape in isolation (react 18.3.1 + vaul 1.1.2, jsdom) with every write to `document.body.style.pointerEvents` logged:

| story shape | write order | final |
|---|---|---|
| `<DrawerProvider open>` (this PR) | `["none", "auto"]` | `"auto"` — no lock, so nothing to catch |
| controlled `open` flipped after mount | `["auto", "none"]` | `"none"` — the actual regression |

The bug requires a **post-mount** controlled flip, because vaul's other two unlock writes live inside `useControllableState.onChange` (`:892-902`), and `onChange` only fires for internally-originated changes — never when a parent flips a controlled `open` (`:485-501`). That's the `controlled.stories.tsx` pattern, and it's the path your hook actually rescues.

Suggested shape — drive it from a button and assert the invariant directly, so a failure names the regression instead of surfacing as an opaque user-event `pointer-events: none` error:

```tsx
export function ModalInsideDrawer() {
  const [drawerOpen, setDrawerOpen] = useState(false)
  // ...
  <Button onClick={() => setDrawerOpen(true)}>Open drawer</Button>
  <DrawerProvider open={drawerOpen} onOpenChange={setDrawerOpen}>
```

```ts
// in play()
await userEvent.click(
  within(canvasElement).getByRole('button', { name: 'Open drawer' })
)
await waitFor(() => expect(document.body.style.pointerEvents).not.toBe('none'))
```

(`waitFor` is exported from `@storybook/test@8.6.15`.)

Worth confirming locally: stash the hook, run `pnpm test:storybook`, and check the story goes red — then green with it. That settles it either way.

packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+    })
+
+    await userEvent.click(openModalButton, { delay: 200 })
+
+    const modal = getByTestId(document.body, 'confirmation-modal')
+    await expect(modal).toBeVisible()
+
+    const confirmButton = getByText(modal as HTMLElement, 'Confirm')
+    await userEvent.click(confirmButton, { delay: 200 })
+    await expect(modal).not.toBeVisible()
```
matheusps: `modal` here is the reference captured on line 31. Two observations:

- Ariakit's `Dialog` never unmounts (no `unmountOnHide`, so `mounted = !unmountOnHide || state.mounted` is always true); it hides via the `hidden` attribute + `display: none`. So all three `getByTestId(document.body, 'confirmation-modal')` calls (31, 39, 43) return the **same** node. You can query once at the top and drop `openModal` plus both `as HTMLElement` casts — `getByTestId` already returns `HTMLElement`.
- More important: `toBeVisible()` also returns `false` for a *detached* node. So if the node ever did get replaced, this assertion would pass vacuously instead of failing. The sibling test defends against exactly this with `withModalRefUpdate` (`confirmation-modal/stories/tests/open-and-close.stories.tsx:24-27`).

Since the node is in fact stable, the cleaner fix is:

```ts
await waitFor(() => expect(modal).not.toBeVisible())
```

which also covers you the day `[data-sl-modal]` gains an exit transition — Ariakit already emits `data-leave`, and the only reason the immediate assert is reliable today is that `modal.css` has no transition at all.

packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const handleClose = () => {
+    setModalOpen(false)
+  }
+
+  return (
+    <DrawerProvider open>
+      <DrawerPopover size="small">
+        <DrawerHeader>
+          <DrawerHeading>Drawer Heading</DrawerHeading>
+          <DrawerDismiss />
```
matheusps: Nit: `DrawerProvider` receives `open` with no `onOpenChange`/`onClose`, so this dismiss routes to `onOpenChange?.(false)` → `undefined` and does nothing. Harmless here (the test never clicks it), but a fixture where the close button is inert is misleading. Wiring `onOpenChange` as suggested on line 59 fixes it incidentally.

packages/shoreline/src/components/drawer/use-non-modal-drawer-unlock.ts
```diff
@@ -0,0 +1,37 @@
...
+    const observer = new MutationObserver(unlock)
+    observer.observe(body, {
+      attributes: true,
+      attributeFilter: ['style'],
+    })
+
+    return () => {
+      cancelAnimationFrame(raf)
+      observer.disconnect()
+      if (body.style.pointerEvents === 'auto') {
```
matheusps: The cleanup infers authorship from whatever value it finds. It's safe today — I traced the two-simultaneous-drawers case (`''` and `'auto'` are computed-identical on `body`, and the surviving instance's observer no-ops on `''`), and Ariakit never touches body pointer-events — but it will also clear an `'auto'` that **vaul itself** wrote (`vaul/dist/index.mjs:901`). Tracking your own write is cheaper to reason about and immune to both:

```ts
let didUnlock = false

const unlock = () => {
  if (body.style.pointerEvents !== 'none') return
  body.style.pointerEvents = 'auto'
  didUnlock = true
}

// ...

return () => {
  cancelAnimationFrame(raf)
  observer.disconnect()
  if (didUnlock && body.style.pointerEvents === 'auto') {
    body.style.pointerEvents = ''
  }
}
```

Non-blocking — the current form works.

packages/shoreline/src/components/drawer/use-non-modal-drawer-unlock.ts
```diff
@@ -0,0 +1,37 @@
...
+    const unlock = () => {
+      if (body.style.pointerEvents === 'none') {
+        body.style.pointerEvents = 'auto'
+      }
+    }
+
+    unlock()
+    const raf = requestAnimationFrame(unlock)
+
+    const observer = new MutationObserver(unlock)
```
matheusps, line 28: Worth documenting the blast radius in the `TODO` above: while the drawer is open, this observer reverts **any** `body { pointer-events: none }` within a microtask — including one that a legitimately-modal nested overlay set for itself.

Shoreline's own `Modal`/`ConfirmationModal` are Ariakit and block via backdrop + tree-hiding (verified: zero `document.body.style.pointerEvents` writes anywhere in `@ariakit/react-core@0.4.23`), so in-house usage is unaffected — which is why the new story passes. But a consumer nesting a **Radix**-based dropdown, select, or alert-dialog inside the drawer would silently lose that layer's outside-blocking. One sentence in the comment saves the next person the trace.

Separate and probably a follow-up rather than this PR: because vaul never forwards `modal`, Radix still applies `trapFocus` and `hideOthers(content)` (`@radix-ui/react-dialog@1.1.1:132-135, 212-215`). So after this fix the background is reachable by **pointer** but still focus-trapped and `aria-hidden` to assistive tech. If "document body not reachable" is meant broadly, that half remains open.

matheusps (changes requested):
## Review summary

Reviewed with three parallel agents (hook correctness / repo conventions / test quality), then verified the load-bearing claims myself against `vaul@1.1.2`, `@radix-ui/react-dismissable-layer@1.1.0` and `@ariakit/react-core@0.4.23`, plus an isolated jsdom reproduction.

**The fix is correct and well-motivated.** vaul 1.1.2 really does render `DialogPrimitive.Root` without forwarding `modal` (`dist/index.mjs:1341-1352`), so Radix's `DismissableLayer` applies `body { pointer-events: none }` (`dist/index.mjs:68-83`). All three of vaul's own unlock writes are one-shot: one `useEffect(…, [modal])` at Root mount (`:1331-1339`) and two inside `useControllableState.onChange` (`:892-902`) — and `onChange` never fires when a *controlled* `open` prop is flipped from outside. A `MutationObserver` is the right shape for that gap, and the `TODO` with both upstream links is good practice.

I checked the usual suspects and they're clean: no SSR hazard (`document` is touched only inside the effect), no leak or stale closure, no observer feedback loop (the callback re-reads the live value and no-ops on `'auto'`, so each `'none'` costs exactly two mutations), and the two-simultaneous-drawers and nested-Ariakit-modal cases are safe. Forcing `'auto'` doesn't break click-outside-to-dismiss either, since vaul already `preventDefault`s `onPointerDownOutside` when `!modal`.

### The one thing I'd change

**The new play test doesn't reproduce the bug it guards.** Details inline on line 59. Measured in jsdom with the story's exact shape:

- `<DrawerProvider open>` (open at mount) → writes `["none","auto"]` → final `auto`. vaul's own mount rAF already wins, so **the test passes on `main` without this hook**.
- controlled `open` flipped after mount → writes `["auto","none"]` → final `none`. That's the real regression, and it's the `controlled.stories.tsx` pattern.

### Also worth knowing (not this PR's fault)

The play test can't run in CI at all: `.github/workflows/pr.yml:43-44` has `# - name: Test Storybook` commented out (since `482a1453d`, 2024-05-03), `test:ci` is `vitest` which doesn't execute play functions, and `chromatic: { disableSnapshot: true }` keeps it out of the only Storybook gate that does run. Since jsdom reproduces both paths above in a few lines, a small Vitest test on the hook under `drawer/tests/` would actually gate this — that's the one addition I'd ask for beyond re-pointing the story.

### Nits

- PR title: repo history leans on scoped types (`fix(charts):`, `feat(docs):`, `feat(code-connect):`), and the scope feeds Lerna's changelog grouping — `fix(drawer): …` would slot in better.
- On CodeRabbit's "stories trifecta" comment: replied in thread — it's citing stale `AGENTS.md`. `stories/tests/` is the correct existing pattern (`button`, `alert`, `confirmation-modal` all use it); there are **zero** `*.play.stories.tsx` / `*.show.stories.tsx` files in the repo.

Everything else — file placement, meta-level `play`, the unexported internal hook, barrel-vs-deep imports, Biome — matches existing convention. I verified Biome passes on the new files with the pinned 1.9.4.

packages/shoreline/src/components/drawer/stories/tests/modal-inside-drawer.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const cancelButton = getByText(openModal as HTMLElement, 'Cancel')
+    await userEvent.click(cancelButton, { delay: 200 })
+    await expect(
+      getByTestId(document.body, 'confirmation-modal')
+    ).not.toBeVisible()
+  },
+  parameters: {
+    chromatic: { disableSnapshot: true },
+  },
+}
```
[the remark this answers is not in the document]
matheusps: This suggestion is based on stale `AGENTS.md` guidance — the "stories trifecta" doesn't exist in this repo. Verified on `main`:

```
$ find packages -name "*.play.stories.tsx" -o -name "*.show.stories.tsx" -o -name "*.examples.stories.tsx"
(0 results)
```

All 102 story files are `<kebab>.stories.tsx`, and `stories/tests/<name>.stories.tsx` is the established pattern for interaction tests — three precedents: `button/stories/tests/click.stories.tsx`, `alert/stories/tests/dismiss.stories.tsx`, `confirmation-modal/stories/tests/open-and-close.stories.tsx`. All three use `title: 'components/<name>/tests'`, meta-level `play`, and `chromatic: { disableSnapshot: true }` — exactly what this PR does. There is no `play.stories.tsx` or `show.stories.tsx` in `drawer/` to move anything into.

A `show.stories.tsx` row also couldn't catch this regression: `pointer-events` is invisible to pixel diffing, so `disableSnapshot: true` is the right call here.

The file placement is correct as-is. (`AGENTS.md` describing a layout the repo doesn't have is worth a separate fix — it also claims an ≥80% Vitest coverage gate that isn't configured anywhere.)

packages/shoreline/src/components/drawer/tests/use-non-modal-drawer-unlock.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    rerender(<DrawerShell open />)
+    await waitFor(() => {
+      expect(document.body.style.pointerEvents).not.toBe('none')
+    })
+
+    rerender(<DrawerShell open={false} />)
+    await waitFor(() => {
+      expect(document.body.style.pointerEvents).not.toBe('auto')
+    })
```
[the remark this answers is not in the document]
matheusps: This is stale — it was written against `9ecf47303`, and `dcae1aa72` ("test(drawer): tighten pointer-events unlock assertions") already applied two of the three proposed edits. At current HEAD:

- line 60 → `expect(document.body.style.pointerEvents).toBe('auto')`
- line 69 → `toBe('auto')`
- line 74 → `toBe('')`

So "Lines 60 and 69 pass when no lock exists" and "Line 74 also passes if the body remains locked as `'none'`" no longer describe the code — `toBe('auto')` fails on `''`, and `toBe('')` fails on `'none'`.

The only unapplied part is seeding `document.body.style.pointerEvents = 'none'` before render, and I'd leave that off deliberately. The lock these tests need is the *real* one: Radix's `DismissableLayer` writes `body { pointer-events: none }` on the controlled open flip. I measured the write sequence for this exact path (react 18.3.1 + vaul 1.1.2, jsdom):

```
controlled `open` flipped after mount → ["auto", "none"] → final "none"
```

That first `"auto"` is vaul's unconditional mount rAF (`vaul/dist/index.mjs:1331-1339`); the `"none"` is Radix, and without this hook nothing clears it. So `toBe('auto')` at line 60/69 genuinely fails without the fix — the tests are real regression guards as written.

Hand-seeding `'none'` would instead assert the hook against a synthetic lock, which passes even if the integration stops locking (e.g. once vaul forwards `modal` and the hook should be deleted per its `TODO`). That trades a real signal for a weaker one.

matheusps (approved):
## Approved ✅

All five inline comments addressed, and the important one is properly fixed.

**The regression test now actually reproduces the regression.** The story drives the drawer open post-mount via `open={drawerOpen} onOpenChange={setDrawerOpen}` and asserts the invariant directly with `waitFor(() => expect(document.body.style.pointerEvents).not.toBe('none'))`. That's the path where the bug lives — measured write order for it is `["auto","none"]` → final `"none"` without the hook, versus `["none","auto"]` → `"auto"` for the old open-at-mount shape that vaul already handled.

**The new Vitest test is the change that gives this real CI protection.** `drawer/tests/use-non-modal-drawer-unlock.test.tsx` runs under `pnpm test:ci`, which is the gate that actually executes — unlike `pnpm test:storybook:ci`, still commented out at `.github/workflows/pr.yml:43-44`. Both cases assert positively (`toBe('auto')`, `toBe('')`), so they fail rather than pass vacuously, and they exercise the real Radix lock instead of a hand-seeded one. Nice addition — thanks for going past what I asked for.

Also confirmed fixed: single `getByTestId` query with the `as HTMLElement` casts dropped, `waitFor` on the close assertions, `didUnlock` gating the cleanup, the blast-radius note in the `TODO`, and `DrawerDismiss` is now functional in the fixture since `onOpenChange` is wired.

CI is green — build, lint, and `test:ci` pass on `dcae1aa72`; Chromatic reports no visual changes, as expected for a `pointer-events` fix.

### Not blocking

- CodeRabbit's comment on the test file is stale (written against `9ecf47303`; two of its three edits are already in `dcae1aa72`). Replied in-thread — no action needed. It also withdrew the earlier "stories trifecta" finding.
- Nit, unchanged from my first pass: PR title could be `fix(drawer): …`. Your second commit uses the scope, but the PR title is what feeds the squash message and Lerna's changelog grouping.
- Follow-up, genuinely out of scope here: because vaul never forwards `modal`, Radix still applies `trapFocus` and `hideOthers(content)`. The background is now reachable by pointer but remains focus-trapped and `aria-hidden` to assistive tech. Worth a separate issue if the intent is full non-modality.
- Separate repo cleanup: `AGENTS.md` documents a story trifecta and an ≥80% Vitest coverage gate that don't exist. It sent CodeRabbit down the wrong path on this PR and will do it again.

